Issue #4, opened 2020-03-06
Add Nextcloud


Ornias1993:
It should be relatively doable to add a nextcloud installer based on this project:

https://github.com/danb35/freenas-iocage-nextcloud

Ornias1993:
PR's to include this can now be PR'ed agaist the jails/nextcloud branch.

Ornias1993:
Done, merged into dev branch

PR #22, opened 2020-03-07
Add Nextcloud

Ornias1993:
This adds Nextcloud to the dev branch.
The current version supports Install and Reinstall and stores both its config, database and files persistent across reinstalls.

Some future improvements should be noted:
- Set database datasets to the right settings (recordsize etc)
- More thorough testing on the update script
- Write our own documentation, as the included documentation doesn't (fully) apply anymore

In contrast to some other issues i've defined "Blocked" due to lack of a seperate database server, I'm inclined to keep the integrated database server in this PR, as this PR is mostly ported from @danb35, is already quite complex and I estimate a lot more people are interested in running nextcloud using its own database server.

Simply put: The database is already there and "just works".


Ornias1993:
This fixes #4

Issue #8, opened 2020-03-06
Add InfluxDB


Ornias1993:
It would be nice adding a solid jail of influxDB to the mix

colemamd:
#45 Pull Request to add InfluxDB

Ornias1993:
@colemamd Thanks for the effort, I assigned you to this Issue and linked the PR for you... :)

PR #45, opened 2020-03-28
Add InfluxDB

colemamd:
Adds influxdb.


colemamd:
#8 Corrects this issue

jails/influxdb/install.sh
```diff
@@ -0,0 +1,67 @@
...
+JAIL_NAME="influxdb"
+JAIL_IP="$(sed 's|\(.*\)/.*|\1|' <<<"${influxdb_ip4_addr}" )"
+INCLUDES_PATH="${SCRIPT_DIR}/jails/influxdb/includes"
+DATABASE=${influxdb_database}
+
+# Make sure DB_PATH is empty -- if not, InfluxDB will choke
+if [ "$(ls -A "/mnt/${global_dataset_config}/${JAIL_NAME}")" ]; then
+	echo "Reinstall of influxdb detected... Continuing"
+	REINSTALL="true"
+fi
```
Ornias1993: While this works fine It might be preferable to check the existance of a file...
This might false-positive with a failed install...

jails/influxdb/install.sh
```diff
@@ -0,0 +1,67 @@
...
+
+iocage exec "${JAIL_NAME}" chown -R 907:907 /var/db/influxdb
```
Ornias1993: I'm totally missing persistance, which is a requirement for inclusion.
Th actual database should be installed in the config directory, like all other jails.

You can either  (for example).
- mount change the config to save the database under /config
or
- mount /config/db to /var/db/influxdb

Till the persistance is in, merging is blocked...

colemamd: Can you elaborate on what you mean by persistence? Are you saying the db should remain even if the jail is destroyed? If so, I can fix that easily.

colemamd: database location is configured via influxdb.conf. Changing that file to indicate global_dataset_config location

config.yml.example
```diff
@@ -6,10 +6,10 @@ global:
    # Iocage root dataset
    iocage: tank/iocage
    # Media library dataset
-   media: tank/media
+   media: tank/Media
    # Parent downloads dataset 
    # (is going to contain: Complete and Incomplete datasets for downloads)
-   downloads: tank/downloads
+   downloads: tank/iocage/download
```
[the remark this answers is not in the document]
colemamd: yeah, oops. missed that cleanup item before I sent the pr.

jails/influxdb/install.sh
colemamd: Did some more testing, there is no adverse affects if there is an existing database, it will begin using the existing data once recreated. I don't see a need to test for an existing db on install, therefore I removed this.

jails/influxdb/includes/influxdb.conf
```diff
@@ -0,0 +1,586 @@
...
+###
+
+[[opentsdb]]
+  # enabled = false
+  # bind-address = ":4242"
+  # database = "opentsdb"
+  # retention-policy = ""
+  # consistency-level = "one"
+  # tls-enabled = false
+  # certificate= "/etc/ssl/influxdb.pem"
```
Ornias1993, line 496: `/config/ssl/influxdb.pem` would be prefered...

jails/influxdb/includes/influxdb.conf
```diff
@@ -0,0 +1,586 @@
...
+
+  # Enables authentication on the /ping, /metrics, and deprecated /status
+  # endpoints. This setting has no effect if auth-enabled is set to false.
+  # ping-auth-enabled = false
+
+  # Determines whether HTTPS is enabled.
+  # https-enabled = false
+
+  # The SSL certificate to use when HTTPS is enabled.
+  # https-certificate = "/etc/ssl/influxdb.pem"
```
Ornias1993, line 308: `/config/ssl/influxdb.pem` would be prefered...

jails/influxdb/includes/influxdb.conf
```diff
@@ -0,0 +1,586 @@
...
+### for your system, but the WAL settings are an advanced configuration. The
+### defaults should work for most systems.
+###
+
+[data]
+  # The directory where the TSM storage engine stores TSM files.
+  dir = "/config/influxdb/database/data"
+
+  # The directory where the TSM storage engine stores WAL files.
+  wal-dir = "/config/influxdb/database/wal"
```
Ornias1993, line 48: Jailman is a solution that is focussed on persistance across reinstalls.
That means the data should never be inside the jail and should be under /config/ instead.

This is what I meant with persistance... Hope that clears it up. 
I think this would fix that issue, did you test it? :)

config.yml.example
```diff
@@ -6,7 +6,7 @@ global:
    # Iocage root dataset
    iocage: tank/iocage
    # Media library dataset
-   media: tank/media
+   media: tank/Media
```
Ornias1993, line 9: I know the case might be considered "wrong" but I would personally prefer not changing the case in the example, in case people copy-paste too much and mess up their config...

jails/influxdb/install.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Initialise defaults
+JAIL_NAME="influxdb"
+JAIL_IP="$(sed 's|\(.*\)/.*|\1|' <<<"${influxdb_ip4_addr}" )"
+INCLUDES_PATH="${SCRIPT_DIR}/jails/influxdb/includes"
+DATABASE=${influxdb_database}
+
+# Install includes fstab
+iocage exec "${JAIL_NAME}" mkdir -p /mnt/includes
+iocage fstab -a "${JAIL_NAME}" "${INCLUDES_PATH}" /mnt/includes nullfs rw 0 0
```
Ornias1993, line 18: If it's just for copying one file, it would be more efficient code wise not to do the whole fstab thing and just CP it from includes into the /iocage/jails/influxdb/root/etcetcetc folder.

You can look at the jackett jail for an easy example ;)

jails/influxdb/install.sh
```diff
@@ -0,0 +1,49 @@
...
+# Copy and edit pre-written config files
+echo "Copying default config file"
+iocage exec "${JAIL_NAME}" cp -f /mnt/includes/influxdb.conf /usr/local/etc/
+
+# Start influxdb and wait for it to startup
+iocage exec "${JAIL_NAME}" service influxd start
+sleep 15
+
+# Create database and restart
+iocage exec "${JAIL_NAME}" "curl -i -XPOST http://localhost:8086/query --data-urlencode 'q=CREATE DATABASE ${DATABASE}'"
```
Ornias1993, line 32: I guess reinstall was not needed, because this just throws "database already exists" or something simiar?

jails/influxdb/install.sh
Ornias1993: I noticed the changes, commented on them to elaborate... 
But yes this was what I meant, sorry for the confusion :)

jails/influxdb/includes/influxdb.conf
colemamd: Good catch. I just copied the default config, didn't notice the ssl settings... This would need to be coordinated when a reverse proxy jail is added, correct?

jails/influxdb/includes/influxdb.conf
colemamd: I did test, influx does not throw an error if you reinstall into the same location, or add a database of a same name. This is why I took that portion out, it doesn't appear to make a difference if there's an existing db or not...

config.yml.example
colemamd: Sorry, leftovers from my install...

jails/influxdb/install.sh
colemamd: No, surprisingly if there is already an existing database the command still works, no error message. I didn't see a need to even test for an existing db, as it didn't throw an error if there was one already.

jails/influxdb/includes/influxdb.conf
colemamd: btw, these settings are currently commented out and would need to be changed by the user anyway. will make the change to help avoid confusion.

jails/influxdb/install.sh
colemamd: I was actually about to do this, but didn't. Will make the change.

jails/influxdb/install.sh
Ornias1993: If there is already one, sure the old data is still in place? 
If so, that would be enough indeed.

Kinda odd it doesn't throw an error though... But well...

jails/influxdb/install.sh
Ornias1993: Happens to the best of us! :)

PR #58, opened 2020-04-06
Add central ports mount for all jails

### LINKED ISSUE #42: Create and mount /portsnap on jail creation
With the inclussion of nextcloud we now have a central /portsnap dataset that includes the ports tree.
this makes it possible to share this to other jails too.

The most clean-and-easy solution would be to just mount this on jail creation to all jails.

Ornias1993:
Adds central ports mount for all jails,
Also does auto fetch, extract and auto update on install

Closes #42

### DIFF
--- a/global.sh
+++ b/global.sh
@@ -87,6 +87,11 @@ else
 	
 	createmount $1 ${global_dataset_config}
 	createmount $1 ${global_dataset_config}/$1 /config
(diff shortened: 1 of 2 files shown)


Ornias1993:
As there aren't many potential problems and risks (as this code is already used in at least one jail and is basically just moved), i'm going to merge this.

PR #61, opened 2020-04-06
Jail-Specific config.yml

### LINKED ISSUE #43: Enable config.yml for seperate jail fixed config
It would be nice if we could give every jail-config its own config.yml file and load those together with the main config.yml

This would enable us to seperate "user facing" config like usernames, ip-settings, passwords etc, from "internal only" config like "pkgs".

It would also allow us to more easily push changes to the pkgs required and upGRADE packages with updates.

Ornias1993:
This draft PR introduces the ability to add a config.yml to every jail, containing non-user config settings like the PKG list.

It's very basic and not fully tested at the moment, but it gives an impression about it.
With this implemented jail maintainers could for example change the required packages with an update, something that isn't possible when the package list is included in the user config like it is now.

Closes #43

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-config.yml
+/config.yml
--- a/config.yml.example
+++ b/config.yml.example
@@ -28,42 +28,15 @@ example: example
   # Jail specific pkgs. 
   # Please use standard space delimited pkg install syntax.
   pkgs: mono
- 
-jackett: jackett
-  pkgs: mono
- 
-radarr: radarr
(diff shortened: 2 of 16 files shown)


Ornias1993:
I might want some testers for this...

Ornias1993:
I'm going to do something drastic:
This is going to be merged into dev and worked from there to force more testing.

It's starting to become annoying that it needs to be kept up-to-date with all recent changes to jails and this also enforces new jail changes are tested on this.

PR #62, opened 2020-04-07
Fixed destroyjails so it can destroy multiple jails. 

thorsteneb:
Minor formatting cleanup around usage code.


thorsteneb:
I noticed a CRLF in a previous PR I issued. Are you editing on Windows? A core.autocrlf true would help in that case; core.autocrlf input for every other OS or even Windows when using LF-aware editors there (VS Code, vim, Notepad++, etc)

Ornias1993:
I'll merge it right away... as it's such a small fix.

Issue #70, opened 2020-04-15
Document jail creation config options


Ornias1993:
It would be nice making a wiki list of all supported jail config options for the config file.
This does not include jail-specific config options, just the general jail-installation parameters.

Jail specific config options (like the custom plex config options) can/should be included in their specific readme files.

Ornias1993:
Pre-release wiki for 1.2 is already online:
https://ornias1993.github.io/jailman/config%20options/

Issue #68, opened 2020-04-10
Document assumed directory layout for included installers


coconutoil:
I think we need to document and stardardise the general media store layout, like
`/mnt/tank1/media/etc`
as the Radarr script will automatically mount the subdirectory "movies":
`createmount radarr ${global_dataset_media}/movies /mnt/movies `

For example my shows are stored in 
`"${global_dataset_media}/tv_shows" `
but the Sonarr script mounts 
`"${global_dataset_media}/shows"`,
so at this stage I think having the basic assumed layout handy in the config script or somewhere would be beneficial so beforehand we can either edit (in config.yml) or restructure our data accordingly. I think long-term this will be essential.

As a side note (me, on autopilot) renamed my config.yml.sample to config.**yaml** and spent an hour trying to troublehsoot why nothing was installing or working when it should be! (with my dyslexia!).. Damn old ruby days ;)

Ornias1993:
> I think we need to document and stardardise the general media store layout

The general media layout is standardised. It's the same in every jail.

> For example my shows are stored in
> "${global_dataset_media}/tv_shows"
> but the Sonarr script mounts
> "${global_dataset_media}/shows",

Yes, this is expected and the layout should infact be documented. Thanks for reporting!

> so at this stage I think having the basic assumed layout handy in the config script

I'm hesitant to bloat the config file any more than it already is. It would be preferable to have it as KISS as possible, to not overload new users with (customisation related) info.

We have a wiki for it, so i'll put in a change for 1.2 to add the layout to the wiki.

PR #76, opened 2020-04-25
Add documentation for config options

Ornias1993:
- Adds documentation for config.yml options (fixes: #70)
- Describes the dataset structure, included auto-generated datasets (fixes: #68)


Ornias1993:
there is no reason not-to merge new documentation. Unless it's wrong, which it isn't.

Issue #54, opened 2020-04-06
Rename plexpass to beta and document it


Ornias1993:
Rename plexpass to beta and Add documentation in the readme that clarifies it is beta code and not required to use Plexpass features. Maybe also add a comment to the example config.

thorsteneb:
Might even consider renaming the variable in the example config from "plexpass" to "beta" to make this behavior abundantly clear

Ornias1993:
uhmm... I personally prefer to stay close as possible to other implementations.

thorsteneb:
I prefer to confuse users as little as possible, and boy does the "plexpass" naming confuse users.

See FreeNAS 11.3 - in the plugin list, they are now calling the plexmedia-plexpass option the "beta" option, for that is what it is.

There used to be a separate package with PlexPass features, hence the naming. That was merged into the base stable package a while ago, and now all that plexmedia-plexpass does is give users beta code. I think people should make that decision very clearly, and not think for a second that they need that code to use PlexPass features.

Ornias1993:
> See FreeNAS 11.3 - in the plugin list, they are now calling the plexmedia-plexpass option the "beta" option, for that is what it is.

If they changed that for the plugin (wasn't the case last time I checked), we should too.

PR #77, opened 2020-04-25
Rename plexpass to beta and document the beta flag

Ornias1993:
Simple fix for #54 
Plexpass is now named beta and the function has been documented in the readme.


Ornias1993:
This is an easy fix, but we need to look if the name for some variables is still using "plexpass" later down the road.

Issue #20, opened 2020-03-07
Add basic Quicksync support to Plex


Ornias1993:
Because we are not limited by the FreeNAS plugin system, it's very well possible to introduce quicksync hardware encoding for Intel gen4-7 CPU's.

This is a rough explaination/tutorial how to do it:
https://github.com/kern2011/Freenas-Quicksync

This is just the jail side of things, it will just enable the jail to support hardware encoding. it WILL NOT include host fixes and thus WILL NOT enable quicksync.

Ornias1993:
It isn't that hard actually, just put this into the script and test it:
https://github.com/kern2011/Freenas-Quicksync

thorsteneb:
This could be fun. There are some caveats.

- Need to somehow detect that an Intel iGPU is present (lspci and grep?), do a quick kldload and verify that /dev/dri comes in
- Need to identify generation and install the right Intel drivers. Exhaustive list could get hard to maintain
- This breaks IPMI, and there is NO way around that. At the very least there needs to be a prominent warning in docs about that, and during install

Ornias1993:
No. thats not how far I was planning to go with this.
It will not detect intel cpu's. It will just inplement a few of the tweaks, not full quicksync support.

So no, it will not do any of the things you list. So please let me worry about this one ;)

Ornias1993:
I called it "basic support" for a reason.
It would include some basic hooks for users to enable support. Not full blown support.

Ornias1993:
With more detail:
This will just be the Jail portion to prepare for quicksync support on the jail side.
Just things like the driver pkg's etc.

Mostly just the things under: "From the Plex jail console:" in the refered guide.
It's not supposed to be fully documented nor supported yet.

Ornias1993:
As it's mostly a personal note for one of my personal pet projects, i'm going to lock this to prevent endless discussions due to me not fully documenting this brainfart.

Ornias1993:
@thorsteneb I just read your changes... It means I only need to add the video-group command on the jail side. Pushing changes now.

Thanks for your work on upstream!

PR #78, opened 2020-04-25
add plex user to video group

Ornias1993:
- add plex user to video group to enable future hardware encoding jail side. (Fixes: #20)

Thanks to @thorsteneb for checking upstream and verifying it isn't required to add drivers. So only the video group add was needed inside the jail. This finishes #20.


Ornias1993:
This oneliner can't really break anything at all. Mergine.

PR #81, opened 2020-04-25
Update nextcloud

Ornias1993:
push upstream fixes from: https://github.com/danb35/freenas-iocage-nextcloud


Ornias1993:
Fixes come directly from upstream. It's unreasonable to go test those 3 lines again, considering the effort that takes.

Should be included with the general 1.2 test run

Issue #48, opened 2020-04-05
Feature: Safer bash


disconn3ct:
As an end-user, I would prefer if the script exited on error instead of continuing to make changes. Additionally, it would be nice to run static analysis (shellcheck)

For example, https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/ documents some settings that can manage errors and even clean up after failures (ERR trap)

Ornias1993:
I totally forgot to add a shellcheck issue, thanks for adding. Yes this is prefered.

If an individual jail throws an error, it should stop creating said jail indeed.
But there are some edge cases to error handling that should also be checked, there are currently some edge cases where a failure during creation requires manual internvention to delete created files (because it would otherwise think it's a reinstall)

Ornias1993:
To give an update:
- Previous version we added shellcheck repo wide
- Currently some code has been refactored to work with stricter bash settings.

To completely finish this would be something for the long(er) term, because it means going over a lot of the code.

Ornias1993:
@asmod3us Could you look at testing some more parts of the code with stricter bash settings?
It's not my specialty honestly...

asmod3us:
Yep, I'll look into this

Ornias1993:
Closed due to being fixed in the dev branch

Issue #38, opened 2020-03-12
Add Basic CI-ish checks for new jails PR's


Ornias1993:
It would be nice if new Jail PR's got verified using github actions on a basic level:

- Config.yml.example should contain an example reference
- jails/$JAILNAME/install.sh should exist
- jails/$JAILNAME/update.sh should exist
- jails/$JAILNAME/README.md should exist

PR #84, opened 2020-04-25
Linter cleanup

Ornias1993:
This PR cleans all current linter errors, or (if they are wrong/irrelevant) ignores them.
It fixes the already added fix for #38
and also adds most of the work needed for #48


Ornias1993:
As any new issues would magically appear during all other work for 1.2, i'll merge it in 24 if no review happens.

Also: This was a pain and I don't want to redo any more PR's.

PR #87, opened 2020-04-26
Fixes for recent changes

Ornias1993:
Some of the recent changes introduced new (high priority) bugs. This WIP is a bugfix PR


Ornias1993:
Due to priority (without this patch nothing works at all), this is going to be merged right away.
It has been tested to at least fix installing the *arr jails

PR #88, opened 2020-04-26
[WIP] Allow multiple installations of same blueprint

### LINKED ISSUE #89: Enhancement: Creation of portsnap directed by jails/JAILNAME/config.yml
Currently, portsnap is created and mounted into a jail by default, whether it is useful there or not.

I suggest to instead, introduce a parameter that can be specified globally (default) and overridden on a per-jail basis, via config.yml, on whether to install portsnap at all.

Likely also useful: A parameter to determine which ports to install, if portsnap is being brought in.

This might all already be in flight and I am just catching an in-between half-finished build.

### LINKED ISSUE #91: Mariadb root password not correctly set on install
It seems under some circumstances, the mariadb root password doesn't actually get set correctly on install. 
Changing this from `UPDATE ` to `ALTER` should fix this issue.

### LINKED ISSUE #93: Silent failure of phpmyadmin install with MariaDB blueprint
When installing a mariadb jail, it might fail to correctly install phpmyadmin without any warning.

This happens when `cert_email` is not set, even though we do not require a cert email, caddy has a bug upstream for ages now that enforces the use of cert_email even without SSL cert.

### LINKED ISSUE #94: Mixed different filetypes for rc.d files.
It seems there are currently 2 ways in-use for rc.d service files:
`*.`
and
`*.rc`

I suggest making sure all of them use `*.rc` so everyone can be sure what the use of said file is.

### LINKED ISSUE #95: Bitwarden: don't require type and rename to db_type
Currently there is a "type" variable that needs to be set.
This should actually have been db_type and not be required in the first place.

It's mostly usefull to use sqlite as db.

### LINKED ISSUE #96: Nextcloud: remove postgresql
We don't currently use postgress, nor do am I expecting to create and maintain a jail for it.

If a maintainer for postgresql support shows up, he/she could figure out himself/herself how to connect it with everying.

besides: It hasn't been tested anyway.

### LINKED ISSUE #97: Nextcloud: remove external database support
We are fully moving to a solution where the databases are setup automatically. It doesn't seem right to maintain quite a lot of code in nextcloud to support external DB's only for nextcloud.

We can leave db_type in, just like with bitwarden, but not do anything with it.

### LINKED ISSUE #98: Nextcloud: Move Certificate selection to single variable
There are currently multiple mutualy-exclusive variables to select a certificate/ssl system.
This should be limited to one, string-type variable and input verification. The current system lead to less ease-of-use

Ornias1993:
This is a big overhaul. It restructures most of the config systems.

* It seperates Jail and Blueprint.
* It introduces config versioning
* It allows multiple installs of the same blueprint (previously known as jail)

The Blueprint:
- The install script included in this project.
- Config settings are included with the blueprint, but can be overriden in config.yml

The Jail:
- The actuall jail under FreeNAS, the physical running thing you see in the GUI.
- config is done by the user, no config is included, only an example is included

Also included are some bug fixes:
- MariaDB root password is now set correctly (Fixes #91 )
- Caddy now uses a default email adress to prevent silent failure during PHPMyAdmin Install (Fixes #93)

And included are some misc. enhancements because I was working on said code anyway and those where small fixes:
- Installation of portsnap is now a per-blueprint setting (fixes #89 )
- InfluxDB doesn't install a database anymore by default (Lets install one when needed)
- Made sure all rc.d files use `*.rc` (Fixes #94 )
- Renamed type to db_type and made it optional in bitwarden (Fixes #95 )
- Removal of postgresql from nextcloud (Fixes #96 )
- Removal of external databases from Nextcloud (Fixes #97 )
- Cleaned up certificate/ssl selection variables (Fixes #98 )

Every jail can only have 1 blueprint, but blueprints can be used on multiple jails.

Config versioning is introduced, because the old-config is not forward-compatible and the new config is not backwards compatible. v1.2.0 is a good moment to introduce this, because we are removing pkgs from the jail config (moving to the blueprint config) anyhow.

All Jails tested working.


Ornias1993:
Some basic jails already working using the new system. such as:
Transmission

Quite some work, testing and cleanup needed to get everything running smoothly.

Ornias1993:
Tested Working under the new system:
- Transmission
- Lidarr
- Sonarr
- Radarr

Ornias1993:
KMS and Plex confirmed working

Ornias1993:
Tautulli and Organizr working

Ornias1993:
Also working:
- MariaDB

Left TODO:
- Bitwarden
- Nextcloud
- Unifi

Ornias1993:
Also working:
- Unifi
- Bitwarden

Left TODO:
- Nextcloud

Ornias1993:
Just one jail left, everything except nextcloud can be reviewed.
Will be closed and merged in 48 hours if no replies come in.

Ornias1993:
All jails tested to be working.

Just one jail left, everything except nextcloud can be reviewed.
Will be closed and merged in 48 hours if no replies come in.

@colemamd I would appreciate if you could check/test the unifi and influxdb side of these changes.

PR #101, opened 2020-05-11
Add transmission+openvpn, qbittorrent+openvpn

disconn3ct:
Globally, this adds a service_extraconf variable for passing arbitrary values to iocage.

Using that capability, this PR adds transvpn (transmission+openvpn) and qbitvpn(qbittorrent+openvpn). The strange tun workaround is needed on my system for some reason, possibly related to other users of tun. (Openvpn cannot automatically allocate one, and tun## below around 100 are unavailable to the jail.)

Comments welcome, especially interested in improvements to the ipfw and vpn config import portions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 config.yml
+openvpn.conf
+openvpn-*
--- a/config.yml.example
+++ b/config.yml.example
@@ -48,7 +48,8 @@ transvpn: transvpn
   pkgs: bash unzip unrar transmission openvpn
   ovpnconf: openvpn.conf
   extraconf: allow_tun=1
-  
+  tundev: tun0
+
 plex: plex
   plexpass: false
   pkgs: plexmediaserver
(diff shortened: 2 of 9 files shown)


colemamd:
New features are going into the dev branch. I recommend rebasing to dev (there are LOTS of changes) and submitting your pr against that.

Ornias1993:
Some feedback from me:
Besides rebasing to the new blueprint-based system, I do think there is a place for VPN support.

I think at least the service_extraconf option, could be great combined with the blueprint system!

However:
I would prever VPN support be an option for the normal torrent applications, not a seperate application. Not because I want to be picky, but it becomes troublesome if we need to create lists and lists of different application combinations.
Just like how we added UnifiPoller as an option to Unifi controller basically.

colemamd:
I am very interested in this as well. Actually, I just got deluge working with OpenVPN in one jail. took a while though...

PR #105, opened 2020-05-22
[WIP] Blueprint Init and Exit refactor

Ornias1993:
# Pull Request Template

### Purpose
This PR aims to restructure the blueprint init and exit (consolidate and simplify them).
Its main goal is to make it easier to start creating blueprints.

#### Automatic local variable creation for blueprints
A big portion of the work will be variable standardisation, both inside and outside the install script and remove the need for exclaimation mark referencing of (sub)variables in install scripts.

Currently the new variable structure is pretty mixed up. The new variable structure will be:
Config -> Install Script
`$Category_$Name_$Content -> $Content`

Example:
`jail_mariadbjail_gateway -> gateway`

Sometimes we grab data about other jails from the config file during install. This is a great way to simplify install processes. While this currently works, it would be nice streamlining this.
I suggest using:
`link_$linkname: dbname`

For example:
`link_mariadbdatabase: mariadbjail`

#### Jail Linking
I debated automatically grabbing values from linked jails, but I've opted against it. It would needlessly make the config and scripts harder to work with.

Instead i'm going to write a function that can be pasted into the install script to both grab the config value, put it into a variable and do any error checking.

Something like:
`linkfetch $linkname $value $value $value $value`

which would put each of the values into:
`link_linkname_value`

#### standardising custom variables
When it comes to jail specific variables, such as credentials, specific settings, this will be documented and standardised.

**I suggest something like:**
For this-blueprint-only variables, just:
$value
(to be reviewed on a case-by-case basis)

**For external content:**
$origin/type_$value

For example, to specify an external mariadb database:
mariadb_database

- [ ] Code is sufficiently commented
- [ ] Code is indented with tabs and not spaces
- [ ] The PR does not bring up any new errors
- [ ] The PR has been tested
- [ ] Any new files are named using lowercase (to avoid issues on case sensitive file systems)


Ornias1993:
Succeeded by #120 due to repo move

PR #119, opened 2020-05-24
devnull all the things, we do our own messages

Ornias1993:
# Pull Request Template
### Purpose
This /dev/nulls the git output from the auto updater.
This became annoying as we do our own verbosity (and some error messages aren't relevant for our usecase)

- [X] The PR has been tested


Ornias1993:
The current repo move leads to a confusing error, emergency merge'ing.

PR #120, opened 2020-05-24
Blueprint Init refactor

Ornias1993:
# Pull Request Template

### Purpose
This PR aims to restructure the blueprint init (consolidate and simplify it).
Its main goal is to make it easier to start creating blueprints.

#### Automatic local variable creation for blueprints
A big portion of the work will be variable standardisation, both inside and outside the install script and remove the need for exclaimation mark referencing of (sub)variables in install scripts.

Currently the new variable structure is pretty mixed up. The new variable structure will be:
Config -> Install Script
`$Category_$Name_$Content -> $Content`

Example:
`jail_mariadbjail_gateway -> gateway`

Sometimes we grab data about other jails from the config file during install. This is a great way to simplify install processes. While this currently works, it would be nice streamlining this.
I suggest using:
`link_$linkname: jailname`

For example:
`link_mariadbdatabase: mariadbjail`

Grabbing the config settings from other jails (for example: the IP of a DB server) is not included in this PR.

#### standardising custom variables
When it comes to jail specific variables, such as credentials, specific settings, this will be documented and standardised.

**I suggest something like:**
For this-blueprint-only variables, just:
$value
(to be reviewed on a case-by-case basis)

**For external content:**
$origin/type_$value

For example, to specify an external mariadb database:
mariadb_database

- [X] The PR has been tested


Ornias1993:
All jails are tested manually to work.

blueprints/mariadb/install.sh
```diff
@@ -1,39 +1,23 @@
...
 
 # Make sure DB_PATH is empty -- if not, MariaDB/PostgreSQL will choke
-# shellcheck disable=SC2154
 if [ "$(ls -A "/mnt/${global_dataset_config}/${1}/db")" ]; then
 	echo "Reinstall of mariadb detected... Continuing"
 	REINSTALL="true"
 fi
 
+iocage exec "${1}" service mariadb stop
+iocage exec "${1}" rm -Rf /var/db/mysql
```
Ornias1993, line 20: needs to be removed

blueprints/mariadb/install.sh
```diff
@@ -1,39 +1,23 @@
...
 # Mount database dataset and set zfs preferences
 iocage exec "${1}" service mysql-server stop
 iocage exec "${1}" rm -Rf /var/db/mysql
```
Ornias1993, line 23: We might want to do this delete thing before every mount...

Ornias1993: Actually thats currently not feasable, it needs to be done on a case-by-case basis.

colemamd:
I like this, vars definitely need standardization. We'll need to make sure everything is documented appropriately as well. Recommend specifying case to use as well.

Ornias1993:
@colemamd the global vars (the ones that are there for every jail) are documented 
Already...
But yes: the others need to be pushed as well to the config of the individual jails and the syntax needs to be added to the contribution guidelines.

Ornias1993:
@colemamd Added the requested documentation.

Ornias1993:
I'm going to merge this, as it's a pain to rebase and it has basically been reviewed already and the review feedback (documentation) included

asmod3us:
`vars` and `reqvqrs` - is that vars=variables as in optional vs reqvars=required variables as in "we need this to set up based on the blueprint"?

Ornias1993:
@asmod3us Yes thats the meaning of reqvars.

vars: variables to be loaded from config.yml
reqvars: of which the following are required before install proceeds

PR #125, opened 2020-05-25
Add custom options for iocage-create to config files

### LINKED ISSUE #124: Add option to add additional custom iocage creation options to Blueprints and Jails
# Issue Template
## Description
Like done in #101 it would be nice to be able to specify custom iocage flags to both blueprints and jails. 

### Detailed feature request 
Those won't be actively supported, but do give people options that are currently unavailable. For example: advanced network configurations like VPN.

We cant "copy" all iocage features ourselves (or we would just become some sort of half-baked-iocage2.0), but this would give us some flexibility on a case-by-case basis.

Ornias1993:
# Pull Request Template
### Purpose
- Adds custom iocage create options to blueprint and jail config files.

### Notes:
Fixes #124

- [x] The PR has been tested


Ornias1993:
The implements a part of the fix from #101 by @disconn3ct

PR #126, opened 2020-05-25
global shellcheck SC2154 fix

Ornias1993:
# Pull Request Template
### Purpose
Shellcheck SC2154 isn't compatible with this repo, because we use a lot of either dynamic or nested variables. Thus it can just-as-well be disabled

### Notes:
Please be aware this also means extra care needs to be taken to prevent referencing unset varaibles.

- [X] The PR has been tested

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,1 +1,1 @@
-disable=SC2034
+disable=SC2034,SC2154


Ornias1993:
As this already has been tested fully in the blueprint init refactor, i'm just going to merge this.
We can always review it with the merge into master.

PR #133, opened 2020-05-26
Make auto-update git aware

### LINKED ISSUE #106: Autodetect git branch for auto updater
# Issue Template
### Detailed Bug Report
Currently we have hit a lot of issues where people forgot to change the git branch.
To fix this we should automatically detect the current branch and update from said branch.

### Steps to Reproduce

1.  Create branch
2.  Run the program
3.  Notice it resets to the previous branch.

Ornias1993:
# Pull Request Template
### Purpose
This fix makes auto-update aware of the current checked-out branch using GIT.
It also checks for the old repository URL, and errors if thats still set.
(I might change this to an override instead)

It then updates to said branch, instead of our own.

### Notes:
Fixes #106

- [X] The PR has been tested

### DIFF
--- a/global.sh
+++ b/global.sh
@@ -24,21 +24,30 @@ parse_yaml() {
 
 # automatic update function
 gitupdate() {
-git remote add upstream https://github.com/jailmanager/jailman.git > /dev/null 2>&1
-echo "checking for updates using Branch: $1"
-git fetch upstream > /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


Ornias1993:
This has already been used for a few days and toroughly tested.
But it is wise to skip the review period to have it reviewed with the current PR's in-waiting

PR #135, opened 2020-05-26
Make sure mariadb could save persistant

### LINKED ISSUE #134: Mariadb doesn't always save data in persistent location
# Issue Template
### Detailed Bug Report
It seems mariadb doesn't always put the data in the correct location. It seems the fstab mount doesn't work right, because the db location is already filled before the fstab entry is executed.

### Steps to Reproduce

1.  Install mariadb
2.  check apps/mariadb/db
3.  see empty folder

Ornias1993:
# Pull Request Template
### Purpose
Mariadb doesn't seem to save persistantly all the time, due to the internal jail folder being empty before mount

### Notes:
Fixes #134 

- [X] The PR has been tested

### DIFF
--- a/blueprints/mariadb/install.sh
+++ b/blueprints/mariadb/install.sh
@@ -35,6 +35,8 @@ if [ "$(ls -A "/mnt/${global_dataset_config}/${1}/db")" ]; then
 fi
 
 # Mount database dataset and set zfs preferences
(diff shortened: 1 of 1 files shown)


Ornias1993:
i'm going to merge this with some haste, because it might influence the performance of other open PR's

PR #136, opened 2020-05-26
exit on dataset creation/mount errors

### LINKED ISSUE #107: If pool doesn't exist, no exit 1
# Issue Template
## Description
Currently if your pool doesn't exist, the script can't create all sorts of datasets and may bug out.

It would be nice checking pool existance

### Steps to Reproduce

1.  Create jail with non-existing pool
2.  Watch log output
3.

Ornias1993:
# Pull Request Template
### Purpose
Currently install continues if the dataset creation and mount commands fail
For example: a non existing pool.

This could give all sorts of issues and thus we should at least stop execution when doing the mounts after jail creation.

### Notes:
Fixes #107 (and other issues)

- [] The PR does not bring up any new errors
- [x] The PR has been tested

### DIFF
--- a/global.sh
+++ b/global.sh
@@ -95,13 +95,13 @@ fi
 rm /tmp/pkg.json
 echo "creating jail config directory"
 # shellcheck disable=SC2154
-createmount "${1}" "${global_dataset_config}"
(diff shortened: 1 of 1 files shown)


Ornias1993:
Finished and tested...

PR #146, opened 2020-05-27
Add Forked_Daapd blueprint

asmod3us:
This PR adds a blueprint for forked-daapd with current ffmpeg.

### Notes:

I'm not quite sure yet what to do in the update script. Can you provide some feedback?

- [X] The PR has been tested

### DIFF
--- a/blueprints/forked_daapd/build-daapd.sh
+++ b/blueprints/forked_daapd/build-daapd.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+
+set -o errexit   # Exit on most errors
+set -o errtrace  # Make sure any error trap is inherited
+set -o nounset   # Disallow expansion of unset variables
+set -o pipefail  # Use last non-zero exit code in a pipeline
+
+trap "ERR during daapd build" ERR
+
+# see config.yml for packages
+
+# this does not work in a jail. is there an alternative?
+# sh -c 'echo "fdesc      /dev/fd fdescfs rw      0       0" >> /etc/fstab'
+# sh -c 'echo "proc       /proc   procfs  rw      0       0" >> /etc/fstab'
+# mount /dev/fd
+# mount /proc
+
+WORKDIR=$(mktemp -d -t daapd)
+CONFIG=/usr/local/etc/forked-daapd.conf
+cd "$WORKDIR"
+
+ENABLE64BIT="--enable-64bit"
+PKG_CONFIG_PATH=/usr/local/lib:/usr/local/lib/pkgconfig:/usr/local/libdata/pkgconfig:/usr/local/ffmpeg/lib/pkgconfig
+export PKG_CONFIG_PATH
+
+wget -c --no-check-certificate https://github.com/antlr/website-antlr3/raw/gh-pages/download/antlr-3.4-complete.jar
+wget -c --no-check-certificate https://github.com/antlr/website-antlr3/raw/gh-pages/download/C/libantlr3c-3.4.tar.gz
+
+install antlr-3.4-complete.jar /usr/local/share/java
+printf "#!/bin/sh
+export CLASSPATH
+CLASSPATH=\$CLASSPATH:/usr/local/share/java/antlr-3.4-complete.jar:/usr/local/share/java
+/usr/local/bin/java org.antlr.Tool \$*
+" > antlr3
+install -m 755 antlr3 /usr/local/bin
+
+tar xzf libantlr3c-3.4.tar.gz
+cd libantlr3c-3.4
+./configure $ENABLE64BIT && gmake -j8 && gmake install
+
+cd "$WORKDIR"
+git clone --depth 1 https://github.com/ejurgensen/forked-daapd.git
+cd forked-daapd
+
+autoreconf -vi
+
+PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/usr/local/ffmpeg/lib/pkgconfig"
+export PKG_CONFIG_PATH
+
+GEN_CFLAGS="-I/usr/include"
+LOCAL_CFLAGS="-I/usr/local/include"
+
+ZLIB_CFLAGS=$GEN_CFLAGS
+ZLIB_LIBS="-lz"
+export ZLIB_CFLAGS ZLIB_LIBS
+
+CONFUSE_CFLAGS=$LOCAL_CFLAGS
+CONFUSE_LIBS="-L/usr/local/lib -lconfuse"
+export CONFUSE_CFLAGS
+export CONFUSE_LIBS
+
+MINIXML_CFLAGS=$LOCAL_CFLAGS
+MINIXML_LIBS="-L/usr/local/lib -lmxml"
+export MINIXML_CFLAGS MINIXML_LIBS
+
+SQLITE3_CFLAGS=$LOCAL_CFLAGS
+SQLITE3_LIBS="-L/usr/local/lib -lsqlite3"
+export SQLITE3_CFLAGS SQLITE3_LIBS
+
+LIBEVENT_CFLAGS=$LOCAL_CFLAGS
+LIBEVENT_LIBS="-L/usr/local/lib -levent"
+export LIBEVENT_CFLAGS LIBEVENT_LIBS
+
+JSON_C_CFLAGS="-I/usr/local/include/json-c"
+JSON_C_LIBS="-L/usr/local/lib -ljson-c"
+export JSON_C_CFLAGS JSON_C_LIBS
+
+LIBAV_CFLAGS="-I/usr/local/ffmpeg/include"
+LIBAV_LIBS="-L/usr/local/ffmpeg/lib -aac -lasound -lavcodec -lavdevice -lavfilter -lavformat -lavutil -lfdk-aac -lpostproc -lswresample -lswscale -lx264 -lx265 -L/usr/local/lib -drm -lX11 -lXau -lXdmcp -lao -lbz2 -lc++ -ldl -lexpat -lfontconfig -lfreetype -lfribidi -lgcc -lgcc_s -liconv -llzma -lm -lpthread -lrt -lva -lvdpau -lxcb -lz -pthread -render -shape -shm -xfixes"
+
+AVAHI_CFLAGS=$LOCAL_CFLAGS
+AVAHI_LIBS="-L/usr/local/lib -lavahi-common -lavahi-client"
+export AVAHI_CFLAGS AVAHI_LIBS
+
+LIBPLIST_CFLAGS=$LOCAL_CFLAGS
+LIBPLIST_LIBS="-L/usr/local/lib -lplist"
+export LIBPLIST_CFLAGS LIBPLIST_LIBS
+
+LIBSODIUM_CFLAGS=$LOCAL_CFLAGS
+LIBSODIUM_LIBS="-L/usr/local/lib -lsodium"
+export LIBSODIUM_CFLAGS LIBSODIUM_LIBS
+
+LIBWEBSOCKETS_CFLAGS=$LOCAL_CFLAGS
+LIBWEBSOCKETS_LIBS="-L/usr/local/lib -lwebsockets"
+export LIBWEBSOCKETS_CFLAGS LIBWEBSOCKETS_LIBS
+
+CFLAGS="-I/usr/local/include"
+LDFLAGS="-L/usr/local/lib -lz -lmxml -ldns_sd -L/usr/local/ffmpeg/lib -lasound -lavcodec -lavdevice -lavfilter -lavformat -lavutil -lfdk-aac -lpostproc -lswresample -lswscale -lx264 -lx265"
+export CFLAGS LDFLAGS
+
+./configure && gmake -j8
+
+if [ -f $CONFIG ]; then
+    echo "Backing up old config file to $CONFIG.bak"
+    cp "$CONFIG" "$CONFIG.bak"
+fi
+gmake install
+sed -i -- 's/\/var\/cache/\/usr\/local\/var\/cache/g' $CONFIG
+
+# Setup user and startup scripts
+echo "daapd::::::forked-daapd:/nonexistent:/usr/sbin/nologin:" | adduser -w no -D -f -
+chown -R daapd:daapd /usr/local/var/cache/forked-daapd
+if [ ! -f scripts/freebsd_start_10.1.sh ]; then
+    echo "Could not find FreeBSD startup script"
+    exit
+fi
+install -m 755 scripts/freebsd_start_10.1.sh /usr/local/etc/rc.d/forked-daapd
+
--- a/blueprints/forked_daapd/install.sh
+++ b/blueprints/forked_daapd/install.sh
@@ -12,8 +12,6 @@ initblueprint "$1"
 
 createmount "$1" "${itunes_media}" /media/itunes
 
-# shellcheck disable=SC2154
-
 cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-ffmpeg.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
 cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-daapd.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
 
--- a/blueprints/forked_daapd/install.sh
+++ b/blueprints/forked_daapd/install.sh
@@ -49,7 +49,4 @@ else
 	JAIL_IP=${ip4_addr%/*}
 fi
 
-# Done!
-echo "Installation complete!"
-echo "forked-daapd is available at http://${JAIL_IP}:3689/ and via daap."
-echo ""
+exitblueprint "$1" "forked-daapd is available at http://${JAIL_IP}:3689/ and via daap."
--- a/blueprints/forked_daapd/install.sh
+++ b/blueprints/forked_daapd/install.sh
@@ -1,4 +1,8 @@
-# shellcheck disable=SC2154
+#!/usr/local/bin/bash
+# This script builds and installs the current release of forked_daapd
+
+initblueprint "$1"
+
 createmount "$1" "${itunes_media}" /media/itunes
 
 cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-ffmpeg.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
--- a/blueprints/forked_daapd/install.sh
+++ b/blueprints/forked_daapd/install.sh
@@ -45,5 +45,5 @@ iocage exec "$1" service forked-daapd start
 
 # Done!
 echo "Installation complete!"
-echo "forked-daapd is available at http://${JAIL_IP}:3689/ and via daap"
+echo "forked-daapd is available at http://${ip4_addr%/*}:3689/ and via daap."
 echo ""
--- a/blueprints/forked_daapd/install.sh
+++ b/blueprints/forked_daapd/install.sh
@@ -1,0 +1,26 @@
+# shellcheck disable=SC2154
+createmount "$1" "${global_dataset_media}" /media
+
+cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-ffmpeg.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
+cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-daapd.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
+
+iocage exec "$1" pkg install -y alsa-lib autoconf automake autotools avahi avahi-libdns bash cmake curl fontconfig fribidi gettext git glib gmake gperf iconv json-c libconfuse libevent libgcrypt libiconv libinotify libplist libsodium libsoxr libtool libunistring libwebsockets mercurial mxml nasm openjdk8-jre opus pkgconf rsync sqlite sqlite3 wget yasm
+
+iocage exec "$1" bash /root/build-ffmpeg.sh
+iocage exec "$1" bash /root/build-daapd.sh
+
+# default config: /usr/local/etc/forked-daapd.conf
+iocage exec "$1" cp /usr/local/etc/forked-daapd.conf /config/
+iocage exec "$1" sed -i '' -e "/directories =/s?=.*?= { \"/media/$global_dataset_paths_itunes\" }?" /config/forked-daapd.conf
+
+iocage exec "$1" sysrc "dbus_enable=YES"
+iocage exec "$1" sysrc "avahi_daemon_enable=YES"
+iocage exec "$1" sysrc "forked_daapd_flags=-c /config/forked-daapd.conf"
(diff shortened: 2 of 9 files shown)


asmod3us:
Another thing I ran into: the installation of packages via `config.yml` seemed to be magnitudes slower than running pkg install from the install script. That might have been a temporary glitch when I was testing, the FreeBSD servers were on modem speed at that time.

blueprints/forked_daapd/build-daapd.sh
```diff
@@ -0,0 +1,118 @@
...
+WORKDIR=$(mktemp -d -t daapd)
+CONFIG=/usr/local/etc/forked-daapd.conf
+cd "$WORKDIR"
+
+ENABLE64BIT="--enable-64bit"
+PKG_CONFIG_PATH=/usr/local/lib:/usr/local/lib/pkgconfig:/usr/local/libdata/pkgconfig:/usr/local/ffmpeg/lib/pkgconfig
+export PKG_CONFIG_PATH
+
+wget -c --no-check-certificate https://github.com/antlr/website-antlr3/raw/gh-pages/download/antlr-3.4-complete.jar
+wget -c --no-check-certificate https://github.com/antlr/website-antlr3/raw/gh-pages/download/C/libantlr3c-3.4.tar.gz
```
Ornias1993, line 27: Do we want this version hardcoded?

blueprints/forked_daapd/install.sh
```diff
@@ -0,0 +1,26 @@
...
+iocage exec "$1" sysrc "avahi_daemon_enable=YES"
+iocage exec "$1" sysrc "forked_daapd_flags=-c /config/forked-daapd.conf"
+iocage exec "$1" sysrc "forked_daapd_enable=YES"
+
+iocage exec "$1" service dbus start
+iocage exec "$1" service avahi-daemon start
+iocage exec "$1" service forked-daapd start
+
+# remove build depdendencies
+iocage exec "$1" pkg delete -y autoconf automake autotools cmake curl git gmake gperf iconv libtool mercurial nasm opus rsync wget yasm
```
Ornias1993, line 26: In preparation of the other PR's, it might be wise to already add an one-line echo instruction where to connect to as a last line.

blueprints/forked_daapd/readme.md
```diff
@@ -0,0 +1,17 @@
...
+your iTunes Library relative to the `media` dataset.
+
+```
+  ...
+  dataset:
+     ...
+    # Media library dataset
+    media: tank/media
+    paths:
+      itunes: Music/iTunes Music
```
Ornias1993, line 15: Nope, No global paths added please.
You can very well add a path to your jail config without any issue.

Ornias1993:
@asmod3us Thanks for your PR.
For the update script: There are some examples already available where the update script doesn't really do anything.

But I guess an update in this gues would be rebuilding the binary?

blueprints/forked_daapd/readme.md
asmod3us: You mean in `/root/config.yml` under the `jail:` map?

blueprints/forked_daapd/install.sh
asmod3us: Good idea, adding!

blueprints/forked_daapd/build-daapd.sh
asmod3us: I think so, took that directly from the [upstream build instructions](https://github.com/ejurgensen/forked-daapd/blob/master/scripts/freebsd_install_11.0.sh).

blueprints/forked_daapd/install.sh
asmod3us: On another note, do we want to add this to the blueprint or jail config in order to run a reverse proxy (traefik/nginx) off of the provided endpoint?

blueprints/forked_daapd/readme.md
Ornias1993: There is only one jail config:
config.yml.example

asmod3us:
> For the update script: There are some examples already available where the update script doesn't really do anything.
> 
> But I guess an update in this gues would be rebuilding the binary?

That's where I was unclear, some examples seem to just restart the service and/or put config into place (transmission). If rebuilding is the intended effect I can add that.

blueprints/forked_daapd/install.sh
Ornias1993: Want to add what?
Traefik endpoints are the job of the person building the traefik blueprint. A preliminary design has been writen down by me a few days back, but thats not a concern for now.

blueprints/forked_daapd/build-daapd.sh
Ornias1993: In that case, why are you linking the 11.0 version here, but are you using a 10.1 sh file in your code?

Ornias1993:
> > For the update script: There are some examples already available where the update script doesn't really do anything.
> > But I guess an update in this gues would be rebuilding the binary?
> 
> That's where I was unclear, some examples seem to just restart the service and/or put config into place (transmission). If rebuilding is the intended effect I can add that.

Hmmm... this is clearly explained in the transmission update file afaik:
`# Transmision is updated during PKG update, this file is mostly just a placeholder`

blueprints/forked_daapd/install.sh
```diff
@@ -0,0 +1,26 @@
+# shellcheck disable=SC2154
+createmount "$1" "${global_dataset_media}" /media
+
+cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-ffmpeg.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
+cp "${SCRIPT_DIR}"/blueprints/forked_daapd/build-daapd.sh /mnt/"${global_dataset_iocage}"/jails/"$1"/root/root/
+
+iocage exec "$1" pkg install -y alsa-lib autoconf automake autotools avahi avahi-libdns bash cmake curl fontconfig fribidi gettext git glib gmake gperf iconv json-c libconfuse libevent libgcrypt libiconv libinotify libplist libsodium libsoxr libtool libunistring libwebsockets mercurial mxml nasm openjdk8-jre opus pkgconf rsync sqlite sqlite3 wget yasm
```
Ornias1993, line 7: I notice some of these are not getting deleted. 
Those should be under `pkg` in config.yml

blueprints/forked_daapd/readme.md
Ornias1993: I checked and you didn't include your jail in the example config anyhow...

Please edit config.yml.example and add your jail.
Please don't use "path" but use `itunes_path:`, `itunes_media:` or `itunes_dir: `, as it is an outside application.

Ornias1993:
@asmod3us
I've gone ahead and applied some fixes:
- You are now PR'ing to the dev branch, as it should
- I rebased your PR to the dev branch and made it compatible with the updated blueprint layout
- I made a start for your update script
- I added the required entry in config.yml.example
- I moved PKG's that aren't getting removed to the config.yml for your blueprint.
- I added you to the codeowners (which would apply on merge)

asmod3us:
Thank for the updates, a lot to process... haven't played with the dev branch yet.

Ornias1993:
@asmod3us 
Most updates on the dev branch are actually just got merged... Yesterday it was mostly still just Master.
Thats why I helped you ahead, it would be silly if I merge something and you need to fix the consequences (something I don't like about some OSS projects tbh).

I wouldn't suggest using dev branch for production or anything sensitive though, but new jails are supposed to go there. Because refactors go into the dev branch, which means someone has to go rewrite all new blueprints that would be merged into master. ;)

asmod3us:
I'm troubleshooting something on the dev branch:
bash -x snippet of it:
```
+ '[' -n '' ']'
+ iocage fstab -a daapd /mnt/tank/apps/portsnap/ports /usr/ports nullfs rw 0 0
Successfully added mount to daapd's fstab
+ '[' '' == true ']'
+ echo 'Ports not enabled for blueprint, skipping'
Ports not enabled for blueprint, skipping
+ echo 'Jail creation completed for daapd'
Jail creation completed for daapd
+ /mnt/tank/sys-new/jailman/blueprints/forked_daapd/install.sh daapd
Set variable itunes_media to /mnt/tank/media/Music/iCloud Music/iTunes Music
environment: line 5: !value: unbound variable
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Update'
No jails to Update
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Upgrade'
No jails to Upgrade
```

any ideas where the environment: ... output comes from?

Ornias1993:
@asmod3us initblueprint
Spaces in variable values are currently not supported, like at all. 

I should document it tbh.

Ornias1993:
You could try changing:
`declare -g "${var}=${!value}"`
Into:
`declare -g "${var}"="${!value}"`

In `global.sh` under `initblueprint`

asmod3us:
I added that but I don't think that's it. Set the value to `/mnt/tank/foo` and getting same error.
Any better way to troubleshoot than `bash -x`?

```
+ '[' '' == true ']'
+ echo 'Ports not enabled for blueprint, skipping'
Ports not enabled for blueprint, skipping
+ echo 'Jail creation completed for daapd'
Jail creation completed for daapd
+ /mnt/tank/sys-new/jailman/blueprints/forked_daapd/install.sh daapd
Set variable itunes_media to /mnt/tank/foo
environment: line 5: !value: unbound variable
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Update'
No jails to Update
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Upgrade'
No jails to Upgrade
```

Ornias1993:
@asmod3us 

This is wrong in your repo:
```
  forked_daapd: 
    blueprint: forked_daapd
    ip4_addr: 192.168.1.252/24
    gateway: 192.168.1.1
	itunes_media: /mnt/itunes
```

The itunes media part had to be spaces, not a tab.
I pushed a fix for the example.

asmod3us:
i think i fixed that already in the version i'm testing. what you're saying is that itunes_media should b under the forked_daapd map on the same level as gateway?

Ornias1993:
@asmod3us yes, it's YAML.

asmod3us:
I think the echo in `initblueprints` loop needs to read:
```
	echo "Set variable $var to ${!value}"
```

but that's also not it ;-)

asmod3us:
Yep, running with
```
  daapd:
    blueprint: forked_daapd
    dhcp: on
    itunes_media: /mnt/tank/foo
```

I added some debug output and it seems to choke on ip4_addr variable not being set.

```
setting itunes_media
Set variable itunes_media to /mnt/tank/foo
setting ip4_addr
environment: line 6: !value: unbound variable
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Update'
No jails to Update
+ '[' 0 -eq 0 ']'
+ echo 'No jails to Upgrade'
No jails to Upgrade
```

What requires that? I'm using dhcp for the jail.

Ornias1993:
@asmod3us DHCP is not supported. Please do not use advanced unsupported configs for testing a new jail.

Ornias1993:
That being said: It shouldn;t have trouble with null values.

What I find most strange, is the fact your scripts stops execution. Are you running any non-standard bash flags?

asmod3us:
Hmm, dhcp was working fine and documented on the master branch... what would it take to add support back in dev?

asmod3us:
Non-standard is a matter of perspective ;-)
I'm using

```
set -o errexit   # Exit on most errors
set -o errtrace  # Make sure any error trap is inherited
set -o nounset   # Disallow expansion of unset variables
set -o pipefail  # Use last non-zero exit code in a pipeline

trap "ERR during forked-daapd install" ERR
```
as a prelude to most scripts i write

Ornias1993:
@asmod3us 
About bash flags:
Okey, now I'm getting annoyed. I work my ass off to support your work and you start testing with all sorts of custom non-standard settings without saying so.

I have a WHOLE other issue open to make the scripts compatible with those bash flags. >.<
The scripts are NOT compatible with ANY bashflags other than those already set.

About DHCP:
DHCP is INCLUDED, it's not SUPPORTED.
(Which is the same as on master, I just quoted the master docs).

Supported means: You can complain about it.
Not-Supported but included means: You can try and use it, but I don't care if it breaks or doesn't work.

Ornias1993:
The scripts are KNOWN to throw errors that can safely be ignored at times, thats why I never enabled more bash flags.

There is an issue in the issue tracker #48 to support safer bash flags. If the flags you used worked without issue, I would've already add more "end on error" type bashflags myself ;)

Ornias1993:
Anyhow:
- I suggest you start testing without those bash flags.

It might work just fine with DHCP, it just isn't supported.

asmod3us:
Sorry to hear you're annoyed. This PR worked fine on master and I'm trying to make it work on dev, so thank you for your support. I prefer my scripts to abort on errors, so I'm used to writing them that way. I'll remove them from the install script for this PR.

Ornias1993:
@asmod3us There might be a way to fix this...

I have the feeling your flags freak out here:
```
value="jail_${1}_$var"
declare -g "${var}=${!value}"
```
I think it freaks out on expanding these variables if the ${!jail_${1}_$var} doesnt exist

You could try this instead:
```
dummy=""
value="jail_${1}_$var"
value="${!value:-$dummy}"
declare -g "${var}"="${value}"
```

It's a bit crude, but should prevent any error due to variable expansion. Because it should just default to dummy afaik.

asmod3us:
Yep, that workaround works. i'm not sure if there's a difference for the variables from global.yml being unset vs being empty.

Another way would be to surround the loop in initblueprint with `set +u` and `set -x`.

Ornias1993:
> Yep, that workaround works. i'm not sure if there's a difference for the variables from global.yml being unset vs being empty.
> 
> Another way would be to surround the loop in initblueprint with `set +u` and `set -x`.

Good to know that works out, you might be able to shrink it a bit... Take a look :)

This one is one of the culprids I think:
`set -o nounset   # Disallow expansion of unset variables`

asmod3us:
Right, yes I meant `set +u` and `set -u`.

Ornias1993:
> Right, yes I meant `set +u` and `set -u`.

Well, I think we both prefer not using overrides if it could be done otherwise... ;)

Ornias1993:
Feel free to include the fix in your Blueprint PR btw.

asmod3us:
Cool, I will do that. `value="${!value:-}"` seems to work or do you prefer a single `set +u` before the loop?

Ornias1993:
if value="${!value:-}" works, i'm totally fine with that :)

I think it's a little more readable than to add set +u, because using set might lead to a scenario where we get more and more set's everywhere...

Ornias1993:
How about even shorter:
```
value="jail_${1}_$var"
declare -g "${var}"="${!value:-$dummy}"
```

blueprints/forked_daapd/install.sh
```diff
@@ -1,10 +1,19 @@
...
+set -o nounset   # Disallow expansion of unset variables
+set -o pipefail  # Use last non-zero exit code in a pipeline
+
+trap "ERR during forked-daapd install" ERR
+
 initblueprint "$1"
 
 createmount "$1" "${itunes_media}" /media/itunes
 
+# shellcheck disable=SC2154
```
Ornias1993, line 15: This isn't needed, SC2154 should be overriden globaly on dev.

blueprints/forked_daapd/install.sh
```diff
@@ -26,5 +43,7 @@ iocage exec "$1" service dbus start
 iocage exec "$1" service avahi-daemon start
 iocage exec "$1" service forked-daapd start
 
-# remove build depdendencies
-iocage exec "$1" pkg delete -y autoconf automake autotools cmake curl git gmake gperf iconv libtool mercurial nasm opus rsync wget yasm
+# Done!
+echo "Installation complete!"
+echo "forked-daapd is available at http://${JAIL_IP}:3689/ and via daap"
```
Ornias1993:
```suggestion
echo "forked-daapd is available at http://${ip4_addr%/*}:3689/ and via daap"
```

includes/global.sh
```diff
@@ -133,7 +133,7 @@ varlist=blueprint_${!blueprint}_vars
 for var in ${!varlist} ${global_jails_vars}
 do
 	value="jail_${1}_$var"
-    declare -g "${var}=${!value}"
+    declare -g "${var}"="${!value}"
 	echo "Set variable $var to ${!var}"
```
Ornias1993: this does not yet include the discussed fix?

asmod3us: Hm strange, 64540b6a69316f2cd2bf424ccbb871361f297111 has them but the diff does not show it.

Ornias1993: Shows now 👍

Ornias1993:
@asmod3us Did you test the shorter version of the fix?
```
value="jail_${1}_$var"
declare -g "${var}"="${!value:-}"
```

asmod3us:
> @asmod3us Did you test the shorter version of the fix?
> 
> ```
> value="jail_${1}_$var"
> declare -g "${var}"="${!value:-}"
> ```

Hang on, that doesn't seem to work.

Ornias1993:
> > @asmod3us Did you test the shorter version of the fix?
> > ```
> > value="jail_${1}_$var"
> > declare -g "${var}"="${!value:-}"
> > ```
> 
> Hang on, that doesn't seem to work.

I already guessed that, but it was worth the try! :)

asmod3us:
Alright, I've ironed that out. One place where I ran into unset variables is checking of parameters.
Here's a proposed change for `createmount` to work with `set -u`: That syntax is a bit ugly, what do you think?

```
@@ -146,7 +146,7 @@ export -f initblueprint
 # $3 = Target mountpoint
 # $4 = fstab prefernces
 createmount() {
-	if [ -z "$2" ] ; then
+	if [ -z "${2+x}" ] ; then
 		echo "ERROR: No Dataset specified to create and/or mount"
 		exit 1
 	else
@@ -157,9 +157,9 @@ createmount() {
 			echo "Dataset already exists, skipping creation of $2"
 		fi
 
-		if [ -n "$1" ] && [ -n "$3" ]; then
+		if [ -n "${1:-}" ] && [ -n "${3:-}" ]; then
 			iocage exec "${1}" mkdir -p "${3}"
-			if [ -n "${4}" ]; then
+			if [ -n "${4:-}" ]; then
 				iocage fstab -a "${1}" /mnt/"${2}" "${3}" "${4}" || exit 1
 			else
 				iocage fstab -a "${1}" /mnt/"${2}" "${3}" nullfs rw 0 0 || exit 1

```

Ornias1993:
@asmod3us 
Thanks for the effort, it had to be done sometime... Just as well might be now...
Anyway:
It might be cleaner to just name the variables in the begining of the script?

Something like this:
```
createmount() {
jail=${1:-}
dataset=${2:-}
mountpoint=${3:-}
fstab=${4:-}
	if [ -z "${dataset}" ] ; then
		echo "ERROR: No Dataset specified to create and/or mount"
		exit 1
	else
		if [ ! -d "/mnt/${dataset}" ]; then
			echo "Dataset does not exist... Creating... ${dataset}"
			zfs create "${dataset}" || exit 1
		else
			echo "Dataset already exists, skipping creation of ${dataset}"
		fi

		if [ -n "${jail}" ] && [ -n "${mountpoint}" ]; then
			iocage exec "${jail}" mkdir -p "${mountpoint}"
			if [ -n "${fstab}" ]; then
				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" "${fstab}" || exit 1
			else
				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" nullfs rw 0 0 || exit 1
			fi
		else
			echo "No Jail Name or Mount target specified, not mounting dataset"
		fi

	fi
}
export -f createmount
```

It isn't even more rows of code, because I removed the commenting above the function...

Ornias1993:
@asmod3us 
I rebased with current dev for you. I also applied the required changes for current dev.

Btw, I see you are looking into defining ip addresses based on dhcp?
Thats a nice fix for the local jail references of ipv4.

We might be able to change dhcp from "unsupported" to "supported, unless specified otherwise". :)

Ornias1993:
@asmod3us 
I implemented both your fixes for stricter use of variables for initblueprint and my own change for createmount in #147 now.

I also simplified grabbing variables of linked jails.

a-staebler:
Great! I've rebased my latest changes. Should we back out the initblueprint rework here if you're adding that in #147?

Ornias1993:
@a-staebler If you could review that one, I could faststrack that merge and we can remove it here, yes.

Ornias1993:
Thanks for the review, I merged those changes into master.
Might still be a few bugs but we'll find out shortly...

asmod3us:
Rebased on latest dev, I'm doing one final test and then I propose we ship it?

Ornias1993:
I'll open it up for review.
If no one has anything to piss on, i'll merge it 24 hours after.

(considering we need to test the full 1.3 packages eventually anyway)

asmod3us:
Great, let's see how it goes! One final thing I've been thinking about: my itunes media is in a subpath in the media dataset. Any ideas on how we would express that for a blueprint?

Example: outside a jail, path to actual itunes library : `/mnt/tank/media/Music/iCloud Music/iTunes Music`. I want to make the blueprint aware that it needs to set up a mount for the media dataset (e.g. `tank/media` -> `/media`), and then find the library at /media/Music/iCloud Music/iTunes Music.

At the same time I want the option to work if a user has his own dataset for the itunes library.
I don't have a good understanding of how iocage works with storage and would appreciate your insights!

Ornias1993:
@asmod3us
Actually jails aren't "dataset-aware"...
createmount is also made that way:
If you already have a folder, with data, it won't try and create a dataset and will just mount the folder.

Simply put: 
jail mounts are just simlinks to folders, regardless if there is a dataset beneat or not.

asmod3us:
Ah, brilliant, so i can just provide the path in dataset syntax to createmount and it'll work!

Ornias1993:
@asmod3us 
yeah I see no way it would not:
```
	else
		if [ ! -d "/mnt/${dataset}" ]; then
			echo "Dataset does not exist... Creating... ${dataset}"
			zfs create "${dataset}" || exit 1
		else
			echo "Dataset already exists, skipping creation of ${dataset}"
		fi

		if [ -n "${jail}" ] && [ -n "${mountpoint}" ]; then
			iocage exec "${jail}" mkdir -p "${mountpoint}"
			if [ -n "${fstab}" ]; then
				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" "${fstab}" || exit 1
			else
				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" nullfs rw 0 0 || exit 1
			fi
		else
			echo "No Jail Name or Mount target specified, not mounting dataset"
		fi
```

asmod3us:
Completed my final test, can confirm it's working!

Ornias1993:
good-good

asmod3us:
> Whoops, forgot to formally approve the Blueprint >.<

🙏

Ornias1993:
Conflict solved, just going to merge it in.
We have enough time to fix any bugs in the dev branch afterwards.

PR #142, opened 2020-05-26
Blueprint exit and includes refactor

Ornias1993:
# Pull Request Template
### Purpose
This PR standardises and refactors the exit portion of the blueprint scripts. Things like: Success messages, connection references and setting the last flags.

This does look like it adds code instead of decreasing the code base. However, this is due to the fact that previously not all blueprints had appropiate success messages and didn't give connection instructions.

Also added is a hook to do future reinstall checks:
It creates a file called "INSTALLED" in the config dir, that we can check to get if a jail is successfully installed.

It also standardises the includes cp behavoir and adds documentation for all functions in global.sh

### Notes:
Fixes #109
and with this and the previous refactors it also Fixes #92

### Please make sure you have followed the self checks below before submitting a PR:

- [X] Code is sufficiently commented
- [X] Code is indented with tabs and not spaces
- [X] The PR does not bring up any new errors
- [X] The PR has been tested
- [X] Any new files are named using lowercase (to avoid issues on case sensitive file systems)

PR #147, opened 2020-05-27
Refactor: Reinstall, CreateMount, Initblueprint

### LINKED ISSUE #113: Standardised reinstall flagging
# Issue Template
### Detailed Bug Report
Currently reinstall is done differently in different blueprints.
It would be nice to standardise it for all blueprints, outside of the blueprints itself.

Suggested solution would be adding a file:
INSTALLED

To every jail after successfull install and check for said file before every install.

Ornias1993:
# Pull Request Template
### Purpose
This PR adds a standardised reinstall check to the initblueprint script.
It's goal is to make sure a blueprint installation has actually finished correctly and to prevent reinstalls triggering after botched installs. So it uses the previously added INSTALLED file added with the blueprint exit refactor.

It also changes Createmount and Initblueprint, to be compatible with stricter bash flags as required by #48 

As an extra it also automatically fetches all variables for linked jails.
From now on one can get the variable for a linked jail by just using link_$jailname_$variable (ex. link_mariadb_ip4_addr)

### Notes:
Blocked because it requires #142 to be merged first
Fixes #113 

- [x] The PR has been tested

### DIFF
--- a/blueprints/example/config.yml
+++ b/blueprints/example/config.yml
@@ -1,5 +1,5 @@
 blueprint:
   example:
     pkgs: nano
-    vars: testvar testvar2 link_mariadb
+    vars: testvar testvar2 link_testjail
     reqvars: testvar
--- a/blueprints/example/config.yml
+++ b/blueprints/example/config.yml
@@ -1,5 +1,5 @@
 blueprint:
   example:
     pkgs: nano
-    vars: testvar testvar2
+    vars: testvar testvar2 link_mariadb
     reqvars: testvar
--- a/blueprints/bitwarden/install.sh
+++ b/blueprints/bitwarden/install.sh
@@ -46,12 +46,12 @@ else
 	openssl req -new -newkey rsa:2048 -days 365 -nodes -x509 -subj "/C=US/ST=Denial/L=Springfield/O=Dis/CN=localhost" -keyout /mnt/"${global_dataset_config}"/"${1}"/ssl/bitwarden-ssl.key -out /mnt/"${global_dataset_config}"/"${1}"/ssl/bitwarden-ssl.crt
 fi
 
-if [ -f "/mnt/${global_dataset_config}/${1}/bitwarden.log" ]; then
+if [ "${reinstall}" = "true" ]; then
 	echo "Reinstall of Bitwarden detected... using existing config and database"
 else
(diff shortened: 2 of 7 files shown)


blueprints/example/config.yml
```diff
@@ -1,5 +1,5 @@
 blueprint:
   example:
     pkgs: nano
-    vars: testvar testvar2
+    vars: testvar testvar2 link_mariadb
```
asmod3us, line 4: This is starting to get very cool, my vision is for this to be come the docker-compose up of jails :)
Just so I understand: `link_mariadb` will provide all variables from the mariadb jail to the install script of this blueprint, right? Do we want an option that allows to be more explicit, e.g. express that we need the `link_mariadb_ip4_addr`?

Ornias1993: yes: link_mariadb would give you all variables, of the jail that is the VALUE of link_mariadb

For example:
`link_mariadb: mymariadbjail `
Will make all variables set for `mymariadbjail` available locally using link_mariadb.
Including `link_mariadb_ip4_addr`. 

The reason it isn't more specific is to keep it simple, while it already is rather complex.

Ornias1993:
I've made the example a little more clear and added documentation for the feature :)

Ornias1993:
It is tested, so i'm just going to merge it and submit fixes as we speak...
That was everyone can get used to it and it would completely finish  #142

asmod3us:
Cool! I am planning to look into auto-wiring jails to nginx/traefik+letsencrypt ssl when daapd is done, this will be very useful!

Ornias1993:
> Cool! I am planning to look into auto-wiring jails to nginx/traefik+letsencrypt ssl when daapd is done, this will be very useful!

Uhmm.. The choice has been made to go traefik instead. I don't think it's wise to maintain both and traefik has dynamic config AND I just finished the first 75% of the config today (see PR)

asmod3us:
Sure, i'll take traefik!

Ornias1993:
Well, you're always free to send in a PR if you don't like it ;) 

Nice thing about Traefik is we don't have to reload config...
Simply put:
We can make it watch a folder and have all jails just drop-in their personal config for their connection and automatically fetches it and sets it up...

asmod3us:
Oh I agree with you, if traefik is that much easier to handle I'll gladly take it over nginx! Plus the automatic handling of letsencrypt and renewals is super nice.

PR #149, opened 2020-05-27
Add Traefik

Ornias1993:
# Pull Request Template
### Purpose
This PR adds traefik as a global reverse proxy.

Some notes:
- It should auto forward http to https (as it would, even without a valid cert) just use a self-signed one
- It should automatically get an A+ SSLlabs score
- Strict SNI should be an option
- Should be able to set lets-encrypt staging
- It should be able to get wildcard certs using letsencrypt
- It should be able to get custom (non wildcard) certs using letsencrypt
- It should be DNS verification compatible, with all supported DNS providers
- I don't give a flying f about lets encrypt http verification, DNS verification is free and the future, get over it.
- It should read a whole dynamic configs folder for jails to put their configs in.
- It should use (seperate) dynamic configs for every jail

### Notes:
It doesn't contain a few aspect that need to be added with a future patch:
- Support for the actual blueprints
- Support for some more (and custom) headers
- Support for basic auth and forward auth

- [X] The PR has been tested

### DIFF
--- a/blueprints/traefik/config.yml
+++ b/blueprints/traefik/config.yml
@@ -1,5 +1,5 @@
 blueprint:
   traefik:
     pkgs: traefik2
-    vars: domain_name cert_email cert_provider cert_wildcard_domain cert_staging
-    reqvars: cert_email cert_provider
+    vars: domain_name cert_email dns_provider cert_wildcard_domain cert_staging cert_strict_sni
+    reqvars: cert_email dns_provider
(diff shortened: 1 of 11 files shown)


Ornias1993:
Basic Traefik instance runs when manually started with:
`traefik --configFile=/config/traefik.toml`

Ornias1993:
@asmod3us
This is the Traefik PR I was talking about.

it already simplified Traefik config to less than 25 config.yml fields, which is quite nice I think.

Ornias1993:
As I already need multiple rounds of additional PR's to this, I'm just going to merge it.

PR #152, opened 2020-05-28
Update issue and PR templates

Ornias1993:
Current templates are not that nicely done and are also part of the wiki. This PR fixes that issue and creates multiple catagories.


Ornias1993:
This is not going as it should. closing

PR #155, opened 2020-05-28
Add sabnzbd blueprint

asmod3us:
- [X] The PR has been tested

### DIFF
--- a/blueprints/sabnzbd/install.sh
+++ b/blueprints/sabnzbd/install.sh
@@ -1,0 +1,36 @@
+#!/usr/local/bin/bash
+
+initblueprint "$1"
+
+# Check if dataset Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}" /mnt/downloads
+
+# Check if dataset Complete Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}"/complete /mnt/downloads/complete
+
+# Check if dataset InComplete Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}"/incomplete /mnt/downloads/incomplete
+
+iocage exec "$1" chown -R _sabnzbd:_sabnzbd /config
+iocage exec "$1" sysrc "sabnzbd_enable=YES"
+iocage exec "$1" sysrc "sabnzbd_conf_dir=/config"
+iocage exec "$1" sysrc "sabnzbd_user=_sabnzbd"
+iocage exec "$1" sysrc "sabnzbd_group=_sabnzbd"
+# start once to let service write default config
+iocage exec "$1" service sabnzbd start
+iocage exec "$1" service sabnzbd stop
+# put our config in place
+iocage exec "$1" sed -i '' -e 's?host = 127.0.0.1?host = 0.0.0.0?g' /config/sabnzbd.ini
+iocage exec "$1" sed -i '' -e 's?download_dir = Downloads/incomplete?download_dir = /mnt/downloads/incomplete?g' /config/sabnzbd.ini
+iocage exec "$1" sed -i '' -e 's?complete_dir = Downloads/complete?complete_dir = /mnt/downloads/complete?g' /config/sabnzbd.ini
+
+iocage exec "$1" service sabnzbd start
+
+JAIL_IP=${ip4_addr:-}
+if [ -z "${JAIL_IP}" ]; then
+	DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
+	JAIL_IP=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')
+else
+	JAIL_IP=${ip4_addr%/*}
+fi
+exitblueprint "$1" "sabnzbd is now accessible at http://${JAIL_IP}:8080/"
--- a/blueprints/sabnzbd/install.sh
+++ b/blueprints/sabnzbd/install.sh
@@ -24,11 +24,4 @@ iocage exec "$1" sed -i '' -e 's?complete_dir = Downloads/complete?complete_dir 
 
 iocage exec "$1" service sabnzbd start
 
-JAIL_IP=${ip4_addr:-}
-if [ -z "${JAIL_IP}" ]; then
-	DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
-	JAIL_IP=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')
-else
-	JAIL_IP=${ip4_addr%/*}
-fi
-exitblueprint "$1" "sabnzbd is now accessible at http://${JAIL_IP}:8080/"
+exitblueprint "$1" "sabnzbd is now accessible at http://${jail_ip}:8080/"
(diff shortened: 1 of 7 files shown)


Ornias1993:
It seems you got the hang of it, glad to see my streamlining of the blueprint creation system being put to good use :)

blueprints/sabnzbd/install.sh
```diff
@@ -0,0 +1,36 @@
...
+initblueprint "$1"
+
+# Check if dataset Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}" /mnt/downloads
+
+# Check if dataset Complete Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}"/complete /mnt/downloads/complete
+
+# Check if dataset InComplete Downloads dataset exist, create if they do not.
+createmount "$1" "${global_dataset_downloads}"/incomplete /mnt/downloads/incomplete
```
Ornias1993, line 12: comments could be squashed imho:
```suggestion
# Check if (in)complete download datasets exist, create if they do not.
createmount "$1" "${global_dataset_downloads}" /mnt/downloads
createmount "$1" "${global_dataset_downloads}"/complete /mnt/downloads/complete
createmount "$1" "${global_dataset_downloads}"/incomplete /mnt/downloads/incomplete
```

blueprints/sabnzbd/install.sh
```diff
@@ -0,0 +1,36 @@
...
+
+iocage exec "$1" chown -R _sabnzbd:_sabnzbd /config
+iocage exec "$1" sysrc "sabnzbd_enable=YES"
```
Ornias1993, line 15: little comment would be nice:
```suggestion

# Setup sabenzb service
iocage exec "$1" chown -R _sabnzbd:_sabnzbd /config
iocage exec "$1" sysrc "sabnzbd_enable=YES"
```

blueprints/sabnzbd/install.sh
```diff
@@ -0,0 +1,36 @@
...
+iocage exec "$1" sysrc "sabnzbd_conf_dir=/config"
+iocage exec "$1" sysrc "sabnzbd_user=_sabnzbd"
+iocage exec "$1" sysrc "sabnzbd_group=_sabnzbd"
+# start once to let service write default config
+iocage exec "$1" service sabnzbd start
+iocage exec "$1" service sabnzbd stop
+# put our config in place
+iocage exec "$1" sed -i '' -e 's?host = 127.0.0.1?host = 0.0.0.0?g' /config/sabnzbd.ini
```
Ornias1993, line 23: Would personally prefer a white line here:
```suggestion
iocage exec "$1" sysrc "sabnzbd_group=_sabnzbd"

# start once to let service write default config
iocage exec "$1" service sabnzbd start
iocage exec "$1" service sabnzbd stop
# put our config in place
iocage exec "$1" sed -i '' -e 's?host = 127.0.0.1?host = 0.0.0.0?g' /config/sabnzbd.ini
```

blueprints/sabnzbd/install.sh
```diff
@@ -0,0 +1,36 @@
...
+
+iocage exec "$1" service sabnzbd start
+
+JAIL_IP=${ip4_addr:-}
+if [ -z "${JAIL_IP}" ]; then
+	DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
+	JAIL_IP=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')
+else
+	JAIL_IP=${ip4_addr%/*}
+fi
```
Ornias1993, line 35: I think it's time to get a modified variant of this into initblueprint tbh.
If it works nicely, it is a nice addition for all jails that don't require a ip4_addr to be set.
(and it would auto apply to them too)

asmod3us: > It seems you got the hang of it, glad to see my streamlining of the blueprint creation system being put to good use :)

Thanks! Yeah, I'm having fun :)

asmod3us: i'll give it a shot. maybe provide it as a global env var JAIL_IP to the install script?

Ornias1993: We could just remove the mask?
Because after jailcreation nothing ever uses the mask again afaik. Because significant networking (only thing that actually requires masks afaik), isn't done in any jail I know of.

We could do it this way:
```
ip4_addr=${ip4_addr%/*}
ip4_mask=${ip4_addr%*/}
if [ -z "${ip4_addr}" ]; then
	DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
	ip4_addr=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')
fi
```

Ornias1993: The above example would also get rid of the need to use `ip4_addr%/*` in blueprints which is nicer imho.

asmod3us: Good idea, how about something like at the end of initblueprint?

```
	jailip4="jail_${1}_ip4_addr"
	ip4val=${!jailip4:-}

	if [ -z "${ip4val}" ]; then
		DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
		ip4val=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')
		declare -g "jail_ip=${ip4val}"
	else
		declare -g "jail_ip=${ip4val%/*}"
	fi
```

then the blueprint install script could use use `jail_ip` as a variable?

asmod3us: I've added this in 69bd576

includes/global.sh
```diff
@@ -126,39 +126,48 @@ echo "Jail creation completed for ${1}"
...
-	echo "No reinstall flag detected, continuing normal install"
-fi
+	jailip4="jail_${1}_ip4_addr"
+	ip4val=${!jailip4:-}
+	if [ -z "${ip4val}" ]; then
+		DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
+		declare -g "jail_ip=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')"
+	else
+		declare -g "jail_ip=${ip4_addr%/*}"
+	fi
```
Ornias1993, line 170: Some suggestions, ip4_addr is already defined at this stage...

```suggestion
	if [ -z "${ip4_addr}" ]; then
		DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
		declare -g "jail_ip=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')"
	else
		declare -g "jail_ip=${ip4_addr%/*}"
	fi
```

asmod3us: ip4_addr is undefined the whole time if using dhcp, that's why i have the extra check. so this works with both dhcp and a defined ip for the jail.

Ornias1993: No it is defined as "".
The for loop runs ALL variables that are listed under "var" or the global "var".
Look:
```
for var in ${!varlist} ${global_jails_vars}
do
	value="jail_${1}_$var"
	val=${!value:-}
	declare -g "${var}=${val}"
```

Ornias1993: It's part of "global_jails_vars", so it is being processed regardless of being set in config.yml.
Thats why your script errored out without `val=${!value:-}`

asmod3us: Ok, clarification: `-z "$var"` tests for undefined OR empty, which is the case when using dhcp.

Ornias1993: So what is the problem with my solution, did you test it and it didn't work?

asmod3us: Ah, yes I think it ran into an issue, but I’d need to test again to refresh my memory.

asmod3us: Can't repro, simplified it!

asmod3us:
Rebased onto latest dev.

PR #156, opened 2020-05-29
Plex rework

### LINKED ISSUE #139: Add further hardware transcoding support
# Issue Template
### Detailed Bug Report
It would be nice putting some work into adding more hardware transcoding support.

A good start would be porting the work by @thorsteneb on: 
https://github.com/danb35/freenas-iocage-plex/commit/83eaea513c44a8ca2317da6c6907269400485f7e

Ornias1993:
# Pull Request Template
### Purpose
This PR ports the hardware transcoding patch by @thorsteneb from  to jailman.
It's designed to be as compact as possible and keeps the GPLv3 copyrighted code in a seperate function.

### Notes:
Needs additional testing
Fixes #139

- [x] Code is indented with tabs and not spaces
- [x] The PR has been tested

### DIFF
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -11,3 +11,4 @@
 /blueprints/forked_daapd/ @asmod3us
 /blueprints/sabnzbd/ @asmod3us
 /blueprints/sabnzbd3/ @asmod3us
+blueprints\plex\includes\hw_transcoding.sh @thorsteneb
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -11,4 +11,4 @@
 /blueprints/forked_daapd/ @asmod3us
 /blueprints/sabnzbd/ @asmod3us
 /blueprints/sabnzbd3/ @asmod3us
-blueprints\plex\includes\hw_transcoding.sh @thorsteneb
+/blueprints/plex/includes/hw_transcoding.sh @thorsteneb
--- a/blueprints/plex/config.yml
+++ b/blueprints/plex/config.yml
@@ -2,4 +2,4 @@ blueprint:
   plex:
     traefik_service_port: 32400
     pkgs: plexmediaserver
-    vars: beta ramdisk
+    vars: beta ramdisk hw_transcode hw_transcode_ruleset ruleset_script
--- a/blueprints/plex/includes/hw_transcoding.sh
+++ b/blueprints/plex/includes/hw_transcoding.sh
@@ -1,28 +1,26 @@
-# $1 = devfs ruleset number
-# $2 = script location
+#!/usr/local/bin/bash
+# This file contains the devfs setup for plex hardware transcoding
+
+# ${1} = devfs ruleset number
+# ${2} = script location
 # Creates script for devfs ruleset and i915kms, causes it to execute on boot, and loads it
-# Be Aware: This script is GPLv3 Licenced.
 createrulesetscript() {
 	
 # Some input checks
-  if [ -z "$1" ] ; then
+  if [ -z "${1}" ] ; then
     echo "ERROR: No plex devfs ruleset number specified. This is an internal script error."
     return 1
   fi
-  if [ -z "$2" ] ; then
+  if [ -z "${2}" ] ; then
 	  echo "ERROR: No plex ruleset script location specified."
     return 1
   fi
-  if [ -z "$(echo ${RELEASE} | grep '12.1')" ] && [ -z "$(echo ${RELEASE} | grep '11.3')" ] ; then
-    echo "This script only knows how to enable hardware transcode in FreeNAS 11.3 and TrueNAS 12.0"
-    return 1
-  fi
   IGPU_MODEL=$(lspci | grep Intel | grep Graphics) 
-  if [ ! -z "${IGPU_MODEL}" ] ; then
-    echo "Found Intel GPU model " ${IGPU_MODEL} ", this bodes well."
-    if [ -z "$(kldstat | grep i915kms.ko)" ] ; then
+  if [ -n "${IGPU_MODEL}" ] ; then
+    echo "Found Intel GPU model ${IGPU_MODEL}, this bodes well."
+    if kldstat | grep -qv i915kms.ko; then
       kldload /boot/modules/i915kms.ko
-      if [ -z "$(kldstat | grep i915kms.ko)" ] ; then
+      if kldstat | grep -qv i915kms.ko; then
         echo "Unable to load driver for Intel iGPU, please verify it is supported in this version of FreeNAS/TrueNAS"
         return 1
       fi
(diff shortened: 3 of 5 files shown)


Ornias1993:
@thorsteneb 
Would you mind if we used createrulesetscript under BSD2-GPL2?
The version you submitted to @danb35 is GPLv3 and it would simplify things licencing wise....

thorsteneb:
Don't mind at all, have at it.

Ornias1993:
@thorsteneb Feel free to test it (cant test it on compatible hardware atm.)
I left your ruleset creation function mostly intact besides making it pass shellcheck and removing 2 checks we don't need (as we don't support anything before 11.3 anyway...)

thorsteneb:
Not sure how to test hw transcode in plex_rework branch, I don't see relevant options in config.yml.example.

I did note that the install.sh tries to add user plex to group media, which fails because there is no group media. What was the intent?

```
# Add plex user to media group for media accessible
iocage exec "$1" pw groupmod -n media -m plex
```

Ornias1993:
> Not sure how to test hw transcode in plex_rework branch, I don't see relevant options in config.yml.example.

GTM basically. (its documented in the included readme)
It's an experimental advanced feature, those are not included in the example. The example is a quick start, that doesn't include advanced features. 
(simply put, A user should be able to just copy-paste it without much hassle)

> I did note that the install.sh tries to add user plex to group media, which fails because there is no group media. What was the intent?

Thats a WIP. Not really for the hardware transcoding.

thorsteneb:
The test you have to verify that i915kms.ko is not loaded does not work. You use this before loading it manually, and then to abort the hardware transcode install if it's still not found. On my machine with i915kms.ko loaded, it triggers both times as "no i915 found".

Here's some output from a machine where i915kms.ko is already loaded:

```
root@freenas[~/jailman/blueprints/plex/includes]# kldstat | grep -qv i915kms.ko
root@freenas[~/jailman/blueprints/plex/includes]# kldstat | grep i915kms.ko
28    1 0xffffffff8351c000 11ff48   i915kms.ko
root@freenas[~/jailman/blueprints/plex/includes]# kldstat | grep i915
28    1 0xffffffff8351c000 11ff48   i915kms.ko
32    1 0xffffffff836b5000 2328     i915_kbl_dmc_ver1_01_bin.ko
33    1 0xffffffff836b8000 357c0    i915_kbl_huc_ver02_00_bin.ko
```

I think what you are looking for is some variant of "if not" and a grep for the presence of i915.

The "-v" gives you all output other than i915 and doesn't tell you whether it's there or not, same result either way.

Ornias1993:
@thorsteneb You do realise that the idea of grep, was yours right? It's not my code, but I needed to clean it because I didn't even pass the most basic shellchecks...
(things like if [ ! -z ] )

That being said: My idea with -v was wrong indeed...

thorsteneb:
In this case actually "if the grep returns a zero value", that was the whole point of that if construct.

I wasn't even aware of shellcheck. I have no quibbles with following some form of coding guidelines, in this case "! grep -q", relying on the return value of grep rather than the returned output. The bang being "not", so again checking for "grep failure".

Ornias1993:
> In this case actually "if the grep returns a zero value", that was the whole point of that if construct.

As I already said: -v was wrong. You don't need to repeat it, already working on a fix.

 "! grep -q" is also wrong according to shellcheck

thorsteneb:
> "! grep -q" is also wrong according to shellcheck

Which is hilarious, as that's the construct shell check suggested when I ran my original code through it on its web interface.

I'll let you work on that.

Ornias1993:
@thorsteneb Already pushed...
The correct inverse of 
`somecommand | grep -q ""something"`
is
`! somecommand | grep -q "something`

Execution order is:
somecommand
grep
! (inverse result)

I was an idiot thinking -v inversed the result of q, it reversed the result BEFORE q.

thorsteneb:
This is working well so far. One more issue, you appear to have an extra space in the jail name in one line.

```
Setting script /root/plex-ruleset.sh to execute on boot
{"id": 4, "type": "SCRIPT", "command": "", "script": "/root/plex-ruleset.sh", "when": "POSTINIT", "enabled": true, "timeout": 10}
Configuring hardware transcode with ruleset 10.
jail ' plex' not found!
```

Ornias1993:
@thorsteneb Shoot, indeed!
Fixed!

Mind if I set you as codeowner for this script btw? It's basically (besides cleanup) your code and you would get a reminder if add you to the list (+write list)

thorsteneb:
Success, hardware transcode is working in the newly created plex jail. Script has been created in its default location, and there's a task to run it post-init. Looking good.

Unrelated to hardware transcode: The install script suggests connecting to https://, which will fail on a newly created jail in the absence of traefik. http:// works.

```
Traefik-connection not enabled... Skipping connecting this jail to traefik
Jail plex using blueprint plex, installed successfully.
 Plex is now accessible at https://192.168.3.99:32400/web/
```

Ornias1993:
@thorsteneb Interesting, because Plex should auto-generate self signed certs.... :s

thorsteneb:
From testing, I think the way this works is that Plex enables secure connections once the server has been "claimed" - linked to a Plex account. That's an educated guess.

thorsteneb:
> Mind if I set you as codeowner for this script btw?

Yeah that's fine. I don't expect major changes until and unless TrueNAS Core 12.x brings in the ability to tie devfs rulesets to jails.

Ornias1993:
@thorsteneb 
noted, will change, squash and merge... thanks for the help (yet again)

About the 12.x changes:
It might (if they don't change what we want them to) be interesting to look into creating a plex plugin and using the plugin way of setting up devfs.... But thats to be seen, I hope they just open up the api enough for us to be more flexible in setting up devfs "our way" :)

thorsteneb:
> About the 12.x changes:
> It might (if they don't change what we want them to) be interesting to look into creating a plex plugin and using the plugin way of setting up devfs.

Like a community plugin. Yeah, that thought crossed my mind. Not that the "plugin way" seems to be terribly well documented, I'm still confused as to what the exact syntax in that dot json is.

"Community plugin" comes with its own set of headaches though, mainly around user expectations and forum support.

docs/CODEOWNERS
```diff
@@ -11,3 +11,4 @@
 /blueprints/forked_daapd/ @asmod3us
 /blueprints/sabnzbd/ @asmod3us
 /blueprints/sabnzbd3/ @asmod3us
+blueprints\plex\includes\hw_transcoding.sh @thorsteneb
```
asmod3us, line 14: / instead of \?

Ornias1993: F INg Windows >.<

Ornias1993:
> > About the 12.x changes:
> > It might (if they don't change what we want them to) be interesting to look into creating a plex plugin and using the plugin way of setting up devfs.
> 
> Like a community plugin. Yeah, that thought crossed my mind. Not that the "plugin way" seems to be terribly well documented, I'm still confused as to what the exact syntax in that dot json is.
> 
> "Community plugin" comes with its own set of headaches though, mainly around user expectations and forum support.

Well, i'm not suggesting an "actual" community plugin... As in: A plugin supplied seperately, with forum support etc. Because one would hit the same issues one would with IX plugins and most features of JailMan for example wouldn't even be supported

More like "Abusing" the plugin system to make it think it's a plugin and apply devfs settings, but actually not being used like a plugin.

A plugin is a normal jail with an install script an update script, additional files to be merged and a config file that are used after update/install. Thats it basically it.

Simply put: If I install an empty "plugin" with devfs correctly set, I could very well execute all the install commands for plex inside of it using iocage exec and as the update script is also empty, it wouldn't do much to the content.

Ornias1993:
I'm going to merge this, it works and all feedback is processed :)

PR #161, opened 2020-05-29
Auto-connect grafana to unifi-poller data

Ornias1993:
# Pull Request Template
### Purpose
This PR replaces the "datasource" option, with a automated option to grab the unifi poller data from Influxdb

### Notes:
I completely removed the datasource option, because external sources can be added using the UI of grafana anyway and we don't actively support config parameters connecting to external services to keep things simple.
(I just removed those options from nextcloud to external mariadb for example)

- [x] The PR has been tested

### DIFF
--- a/blueprints/grafana/install.sh
+++ b/blueprints/grafana/install.sh
@@ -20,12 +20,16 @@ iocage exec "${1}" chown -R grafana:grafana /config
 
 # Setup connection to influxdb, if necessary
 if [ -n "${link_influxdb}" ]; then
-  cp "${includes_dir}"/influxdb.yaml /mnt/"${global_dataset_config}"/"${1}"/provisioning/datasources
-  iocage exec "${1}" sed -i '' "s|datasource_name|${datasource_name}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|influxdb_ip|${link_influxdb_jail_ip}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|datasource_db|${datasource_database}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|datasource_user|${datasource_user}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|datasource_pass|${datasource_password}|" /config/provisioning/datasources/influxdb.yaml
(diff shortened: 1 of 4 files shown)


Ornias1993:
Besides not being able to find the database, the autofetching of unifipoller credentials has been tested successfully.

blueprints/grafana/install.sh
```diff
@@ -20,12 +20,16 @@ iocage exec "${1}" chown -R grafana:grafana /config
...
-  iocage exec "${1}" sed -i '' "s|datasource_db|${datasource_database}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|datasource_user|${datasource_user}|" /config/provisioning/datasources/influxdb.yaml
-  iocage exec "${1}" sed -i '' "s|datasource_pass|${datasource_password}|" /config/provisioning/datasources/influxdb.yaml
+	if [ -n "${link_unifi}" ]; then
+		cp "${includes_dir}"/influxdb.yaml /mnt/"${global_dataset_config}"/"${1}"/provisioning/datasources/unifi.yaml
+		iocage exec "${1}" sed -i '' "s|datasource_name|unifi|" /config/provisioning/datasources/unifi.yaml
+		iocage exec "${1}" sed -i '' "s|influxdb_ip|${link_influxdb_ip4_addr%/*}|" /config/provisioning/datasources/unifi.yaml
+		iocage exec "${1}" sed -i '' "s|datasource_db|${link_unifi_influxdb_database:-$link_unifi}|" /config/provisioning/datasources/unifi.yaml
+		iocage exec "${1}" sed -i '' "s|datasource_user|${link_unifi_influxdb_user:-$link_unifi}|" /config/provisioning/datasources/unifi.yaml
+		iocage exec "${1}" sed -i '' "s|datasource_pass|${link_unifi_influxdb_password:-}|" /config/provisioning/datasources/unifi.yaml
```
colemamd, line 29: Do we want to allow no password? maybe ${link_unifi_influxdb_password:?"Password cannot be empty"}. Or something similar?

Ornias1993: That should be catches when setting up unifi, grafana would just not connect.. which isnt really a problem imo, because there isnt a valid influxdb database to begin with ;)

Ornias1993: Simply put: If there is a not-working influx password, it would not break anything. it just won't connect which isn't really a problem in the long run.

PR #160, opened 2020-05-29
Fixup *arr versioning and updating

Ornias1993:
# Pull Request Template
### Purpose
Currently some of the *arr jails either don't update using the update script and/or don't auotmatically install using the newest version.

We should streamline this, as one of the selling points of Jailman, In contrast to plugins, is NOT relying on weither the plugin/blueprint version is up-to-date

### Notes:
Fixes #129
Fixes #128
Fixes #127

### Please make sure you have followed the self checks below before submitting a PR:

- [X] Code is sufficiently commented
- [X] Code is indented with tabs and not spaces
- [X] The PR does not bring up any new errors
- [x] The PR has been tested
- [X] Any new files are named using lowercase (to avoid issues on case sensitive file systems)

PR #164, opened 2020-05-30
Add SABnzbd v3.

asmod3us:
- [X] The PR has been tested


Ornias1993:
Just for the sake of people like me that aren't up to speed on the area of NZB clients mind describing the difference between this V3 and the previous one?

Ornias1993:
Is the data structure of V2 compatible with V3?
If so, how similar is the install process?

asmod3us:
> Just for the sake of people like me that aren't up to speed on the area of NZB clients mind describing the difference between this V3 and the previous one?
> 
> _edit_
> I'm stupid and could've just read the first 3 lines of code:
> "Beta1"

It's a complete rewrite in Python 3, and there's not bsd package or port I'm aware of, so the install process is totally different.

Ornias1993:
@asmod3us Okey, Normally i'm not that much a fan of giving new versions of something new blueprints.

But this is precisely one of the exceptions, good catch and thank you for your effort :)

blueprints/sabnzbd3/includes/sabnzbd3.rc
```diff
@@ -0,0 +1,54 @@
...
+. /etc/rc.subr
+
+name=sabnzbd
+rcvar=sabnzbd_enable
+load_rc_config ${name}
+
+: ${sabnzbd_enable:=NO}
+: ${sabnzbd_user:=sabnzbd}
+: ${sabnzbd_group:=sabnzbd}
+: ${sabnzbd_conf_dir="/config"}
```
Ornias1993, line 33: You could do this, you can also keep this stock (if there is a stock rc script) and add it using sysrc...

Your choice in the end.

blueprints/sabnzbd3/install.sh
```diff
@@ -0,0 +1,42 @@
...
+createmount "$1" "${global_dataset_downloads}" /mnt/downloads
+createmount "$1" "${global_dataset_downloads}"/complete /mnt/downloads/complete
+createmount "$1" "${global_dataset_downloads}"/incomplete /mnt/downloads/incomplete
+
+#source_version=3.0.0Beta1
+archive=${blueprint_sabnzbd3_source_version}.tar.gz
+target=/usr/local/share/sabnzbd3
+
+iocage exec "$1" mkdir -p $target
+iocage exec "$1" "curl -qsL https://github.com/sabnzbd/sabnzbd/archive/$archive | tar -xzf - --strip-components 1 -C $target"
```
Ornias1993, line 22: You could do something like I did here, to grab the latest prerelease: #160

blueprints/sabnzbd3/install.sh
```diff
@@ -0,0 +1,42 @@
...
+cp "${includes_dir}"/build.sh "${jail_root}"/root/
+iocage exec "$1" bash /root/build.sh $target
```
Ornias1993, line 24: Is it reallt usefull to seperate 6 lines of code into its own sh script?

blueprints/sabnzbd3/readme.md
```diff
@@ -0,0 +1 @@
+Blueprint for SABnzbd binary newsreader
```
Ornias1993, line 1: First line needs to start with the title that shows up in the wiki. In this case:
```suggestion
# SABnzbd V3
Blueprint for version  3 of the SABnzbd binary newsreader
```

config.yml.example
```diff
@@ -116,7 +115,7 @@ jail:
 
   forked_daapd:
     blueprint: forked_daapd
-    ip4_addr: 192.168.1.252/24
+    ip4_addr: 192.168.1.99/24
```
Ornias1993, line 118: Yeah, I should rework this and just number them top to bottom starting with 1.

includes/global.sh
```diff
@@ -148,6 +148,8 @@ initblueprint() {
 		fi
 	done
 
+	declare -g "jail_root=/mnt/${global_dataset_iocage}/jails/$1/root"
+	declare -g "blueprint_dir=${SCRIPT_DIR}/blueprints/${!blueprint}"
```
Ornias1993, line 152: This should've been done long ago. Thank you!

blueprints/sabnzbd3/includes/sabnzbd3.rc
asmod3us: There isn't a stock rc script, i grabbed the one from v2 and added it to the blueprint.

blueprints/sabnzbd3/install.sh
asmod3us: i did not realize we want "latest" version automatically, I can modify the download to do that.

blueprints/sabnzbd3/install.sh
asmod3us: it could be re-used by an implementation of the update script

blueprints/sabnzbd3/includes/sabnzbd3.rc
Ornias1993: Does that work without issue? :s

blueprints/sabnzbd3/install.sh
Ornias1993: The idea is (for all jails) not to make the same mistakes IX made... So not making users depend on us updating shit, because that never works out the way we want ;)

blueprints/sabnzbd3/install.sh
Ornias1993: Good point, but that goes for about 75% of every update script (because they are mostly repeating a subset of steps from the install scripts anyway)

blueprints/sabnzbd3/includes/sabnzbd3.rc
asmod3us: It works great, I've tested it!

blueprints/sabnzbd3/install.sh
asmod3us: Fair point, I'll change it to pull latest

blueprints/sabnzbd3/install.sh
asmod3us: My motivation is to avoid code duplication, less stuff on my radar! If install and update execute (some) shared code it should be in one common place imho.

blueprints/sabnzbd3/config.yml
```diff
@@ -0,0 +1,3 @@
+blueprint:
+  sabnzbd3:
+    pkgs: git python3 py37-pip py37-sqlite3 unrar par2cmdline-tbb p7zip
```
Ornias1993, line 3: Mind checking if we can use py38 already?

asmod3us: Sure I'll have a look

asmod3us: i cannot see python 3.8 available in pkg at all, default v3 version on freebsd 11.3 seems to be v3.7

asmod3us:
rebased onto dev, ready now.

asmod3us:
> Nice work, some slight notes (considering it's still pretty much WIP)
> Dont forget to add yourself to /docs/codeowners for this blueprint ;)

Done!

blueprints/sabnzbd3/config.yml
Ornias1993: ahh okey, well at least 3.7 is still an actively supported release :)

blueprints/sabnzbd3/update.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/local/bin/bash
+
+initblueprint "$1"
+#TODO insert code to update itself here
```
Ornias1993, line 4: still missing update script

asmod3us: true 🤦

Ornias1993: Well, you wheren't the one who spend 3 hours yesterday realising I need to sanitise URLS being put into SED because SED hates `&` :P

asmod3us:
Fixed a few bugs in update procedure itself while implementing it!

Issue #13, opened 2020-03-06
Cancelled jail creation might require manual intervention


Ornias1993:
Manual intervention might be required if someone cancels jail creation, FreeNAS would request manual destroy a snapshot of the previous creation attempt.

It would be nice to detect this ourselves or at least try removing the snapshot on failure and retrying creation another time.

Ornias1993:
Example of the error in question:
```
Snapshot: tank/iocage/releases/11.3-RELEASE/root@lidarr exists!
Please manually run zfs destroy tank/iocage/releases/11.3-RELEASE/root@lidarr if you wish to destroy it.
Failed to create jail
```

Ornias1993:
Basically we need to catch this error and just run the snapshot destroy command.
(we don't care about snapshotting throw-away jails anyway)

asmod3us:
i think i ran into that recently and passing `-fR` to iocage destroy would get this sorted?

asmod3us:
So the scenario would be an install fails and this message is encountered when attempting a reinstall with `-r`? I will add `-fR` to both our `iocage destroy` calls

Ornias1993:
@asmod3us Not -r only, also just -i with when iocage create failed beforehand.
You sure you want this:
`[-r | --release]	  Destroy a specified RELEASE dataset.`

asmod3us:
any steps to reproduce such an issue? looks like we we would have to parse iocage output and clean up that release dataset?

Ornias1993:
Afaik:
1. Start iocage create
2. hold ctrl+c

asmod3us:
I'm getting this when I do that:
```
❯ iocage list
JAIL is missing it's configuration, please destroy this jail and recreate it.
+-----+--------------+---------+--------------+------+
| -   | sabnzbd3     | CORRUPT | N/A          | N/A  |
+-----+--------------+---------+--------------+------+
```

And a regular `iocage destroy JAIL` got rid of it again, so the above error looks like a different situation.

Ornias1993:
Well at least it now just stops running after this error, thats something :)

asmod3us:
Yes, that is true. And the error message tells you the jail already exists. Then the user can intervene...

Ornias1993:
> Yes, that is true. And the error message tells you the jail already exists. Then the user can intervene...

Well, it isn't the jail that's existing it's a snapshot...
But yes, it literally tells you what you should copy-paste....

[...]

PR #165, opened 2020-05-30
Stricter bash options.

asmod3us:


Ornias1993:
Wow thats a lot, thank you!

You might also want to look at the work @disconn3ct already did on cleaning things up a bit:
https://github.com/disconn3ct/jailman/commit/98d293ffed675af40b3710600e647d3f2deafe61
https://github.com/disconn3ct/jailman/commit/6d8180db3ccb5ec6b99fe9ff9962be7804efccb5

Ornias1993:
On another note, i'm going to look at cleaning sed inputs a bit.
Hit the snag with traefik today (almost finished btw), but:
This works:
`sed -i '' "s|placeholderauthforward|$(echo ${traefik_auth_forward} | sed -e 's/[&]/\\&/g')|"`

while this (obiously) doesnt:
`sed -i '' "s|placeholderauthforward|${traefik_auth_forward}|"`

Why? Because url's (which ${traefik_auth_forward} contains) might contain `&`
So i'm going to sanitise all of those input later on :)

includes/global.sh
```diff
@@ -24,19 +24,20 @@ parse_yaml() {
 
 # automatic update function
 gitupdate() {
+BRANCH=${1:-}
```
Ornias1993, line 27: We could just move the branch detection from jailman.sh here, and get rid of the ${1} entirely...

jailman.sh
```diff
@@ -1,5 +1,72 @@
...
+	done
+
+	echo "Exiting with status ${code}" >&2
+	exit "${code}"
+}
+
+# trap ERR to provide an error handler whenever a command exits nonzero
+#  this is a more verbose version of set -o errexit
+trap 'errexit' ERR
+
```
Ornias1993, line 69: Fancy :)

jailman.sh
```diff
@@ -33,17 +100,27 @@ fi
...
 # Go through the options and put the jails requested in an array
 unset -v sub
+args=("$@")
+arglen=${#args[@]}
+
+installjails=()
+redojails=()
+updatejails=()
+destroyjails=()
+upgradejails=()
```
Ornias1993, line 110: Good catch, we should indeed be defining more variables before using them...

jailman.sh
```diff
@@ -33,17 +100,27 @@ fi
...
 while getopts ":i:r:u:d:g:h" opt
    do
    #Shellcheck on wordsplitting will be disabled. Wordsplitting can't happen, because it's already split using OPTIND.
      case $opt in
         i ) installjails=("$OPTARG")
 			# shellcheck disable=SC2046
-            until [[ $(eval "echo \${$OPTIND}") =~ ^-.* ]] || [ -z $(eval "echo \${$OPTIND}") ]; do
+			until (( OPTIND > arglen )) || [[ $(eval "echo \${$OPTIND}") =~ ^-.* ]]; do
+				echo "at arg $OPTIND"
```
Ornias1993, line 118: Every eval we can get rid off is a win in my book...
Overuse of eval is already WAY to frequent....

jailman.sh
```diff
@@ -33,17 +100,27 @@ fi
...
 				# shellcheck disable=SC2207
-                installjails+=($(eval "echo \${$OPTIND}"))
-                OPTIND=$((OPTIND + 1))
-            done
+				#installjails+=($(eval "echo \${$OPTIND}"))
```
Ornias1993, line 120: This needs to be removed right (the comment)

includes/global.sh
```diff
@@ -7,7 +7,7 @@
 # This function is very picky and complex. Ignore with shellcheck for now.
 # shellcheck disable=SC2086,SC2155
 parse_yaml() {
-   local prefix=${2}
+   local prefix=${2:-}
    local s='[[:space:]]*' w='[a-zA-Z0-9_]*' fs=$(echo @|tr @ '\034')
    sed -ne "s|^\($s\)\($w\)$s:$s\"\(.*\)\"$s\$|\1$fs\2$fs\3|p" \
         -e "s|^\($s\)\($w\)$s:$s\(.*\)$s\$|\1$fs\2$fs\3|p"  "${1}" |
```
Ornias1993, line 13: This whole function is actually an issue, because it doesn't cover lists for example.
Thats beside the fact it covers a prefix that isn't actually used.

Anyway you can rip out the prefix completely for all I care, because we aren't using any of it...

jailman.sh
asmod3us: Yep, I will remove&rebase once it's in a somewhat usable state. Lot's of WIP commits in here right now!

includes/global.sh
asmod3us: > This whole function is actually an issue, because it doesn't cover lists for example.
> Thats beside the fact it covers a prefix that isn't actually used.
> 
> Anyway you can rip out the prefix completely for all I care, because we aren't using any of it...

Ah, ok - out it goes :) I could not understand what prefix was for anyway!

jailman.sh
asmod3us: Getting rid of the second one now!

includes/global.sh
Ornias1993: Its a function I fetched from elsewhere...
TLDR it puts a prefix in front of the resulting env-variable...

asmod3us:
> Wow thats a lot, thank you!
> 
> You might also want to look at the work @disconn3ct already did on cleaning things up a bit:
> [disconn3ct@98d293f](https://github.com/disconn3ct/jailman/commit/98d293ffed675af40b3710600e647d3f2deafe61)
> [disconn3ct@6d8180d](https://github.com/disconn3ct/jailman/commit/6d8180db3ccb5ec6b99fe9ff9962be7804efccb5)

Oh, nice. That's not on dev is it?

Ornias1993:
> > Wow thats a lot, thank you!
> > You might also want to look at the work @disconn3ct already did on cleaning things up a bit:
> > [disconn3ct@98d293f](https://github.com/disconn3ct/jailman/commit/98d293ffed675af40b3710600e647d3f2deafe61)
> > [disconn3ct@6d8180d](https://github.com/disconn3ct/jailman/commit/6d8180db3ccb5ec6b99fe9ff9962be7804efccb5)
> 
> Oh, nice. That's not on dev is it?

It's his own fork... but afaik it's on master, yes... 
A lot would still apply though...

asmod3us:
This is shaping up, I tested an install & destroy and this runs through.

includes/global.sh
asmod3us: implemented in latest bunch of commits

Ornias1993:
@asmod3us Nice work!
Just to be clear: I'm always open for more tests.

Eventually I also want to get good CI testing working.
I've a design prototype in mind (cirrus CI, with FreeBSD) but it needs some work still I hope to have that in with 1.4

jailman.sh
```diff
@@ -15,131 +111,120 @@ source "${SCRIPT_DIR}/includes/global.sh"
...
 	exit 1
 fi
 
 # Check and Execute requested jail destructions
-if [ ${#destroyjails[@]} -eq 0 ]; then 
-	echo "No jails to destroy"
-else
+if [ ${#destroyjails[@]} -gt 0 ]; then
 	# shellcheck disable=SC2124,SC2145
 	echo "jails to destroy ${destroyjails[@]}"
```
Ornias1993, line 210: This should be fixable with:
```suggestion
	echo "jails to destroy " "${destroyjails[@]}"
```

According to:
https://github.com/koalaman/shellcheck/wiki/SC2145

jailman.sh
```diff
@@ -15,131 +111,120 @@ source "${SCRIPT_DIR}/includes/global.sh"
...
 fi
 
 # Check and Execute requested jail Installs
-if [ ${#installjails[@]} -eq 0 ]; then 
-	echo "No jails to install"
-else
-	# shellcheck disable=SC2124,SC2145
+if [ ${#installjails[@]} -gt 0 ]; then
+	# shellcheck disable=SC2145
 	echo "jails to install ${installjails[@]}"
```
Ornias1993, line 222: same:
```suggestion
	# shellcheck disable=SC2145
	echo "jails to install " "${installjails[@]}"
```

jailman.sh
```diff
@@ -155,15 +240,13 @@ else
 fi
 
 # Check and Execute requested jail Reinstalls
-if [ ${#redojails[@]} -eq 0 ]; then 
-	echo "No jails to ReInstall"
-else
-	# shellcheck disable=SC2124,SC2145
+if [ ${#redojails[@]} -gt 0 ]; then
+	# shellcheck disable=SC2145
 	echo "jails to reinstall ${redojails[@]}"
```
Ornias1993, line 245: same:
```suggestion
	# shellcheck disable=SC2145
	echo "jails to reinstall " "${redojails[@]}"
```

jailman.sh
```diff
@@ -178,17 +261,14 @@ else
...
 
-
 # Check and Execute requested jail Updates
-if [ ${#updatejails[@]} -eq 0 ]; then 
-	echo "No jails to Update"
-else
-	# shellcheck disable=SC2124,SC2145
+if [ ${#updatejails[@]} -gt 0 ]; then
+	# shellcheck disable=SC2145
 	echo "jails to update ${updatejails[@]}"
```
Ornias1993, line 267: same:
```suggestion
	# shellcheck disable=SC2145
	echo "jails to update " "${updatejails[@]}"
```

jailman.sh
```diff
@@ -207,15 +287,13 @@ else
 fi
 
 # Check and Execute requested jail Upgrades
-if [ ${#upgradejails[@]} -eq 0 ]; then 
-	echo "No jails to Upgrade"
-else
-	# shellcheck disable=SC2124,SC2145
+if [ ${#upgradejails[@]} -gt 0 ]; then
+	# shellcheck disable=SC2145
 	echo "jails to update ${upgradejails[@]}"
```
Ornias1993, line 292: same:
```suggestion
	# shellcheck disable=SC2145
	echo "jails to update " "${upgradejails[@]}"
```

jailman.sh
```diff
@@ -15,131 +111,120 @@ source "${SCRIPT_DIR}/includes/global.sh"
...
 # auto detect iocage install location
-global_dataset_iocage=$(zfs get -H -o value mountpoint $(iocage get -p)/iocage)
+global_dataset_iocage=$(zfs get -H -o value mountpoint "$(iocage get -p)"/iocage)
 global_dataset_iocage=${global_dataset_iocage#/mnt/}
 export global_dataset_iocage
 
 # Parse the Config YAML
+for configpath in "${SCRIPT_DIR}"/blueprints/*/config.yml; do
+	# shellcheck disable=SC2046
+	! eval $(parse_yaml "${configpath}")
```
Ornias1993, line 193: Errors here are correctly fatal now?

jailman.sh
```diff
@@ -1,8 +1,104 @@
...
+	echo "    Update the jail and any packages inside"
+	echo ""
+	echo " Examples:"
+	echo ""
+	echo "    # $0 -i plex"
+	echo "      Install plex"
+	echo ""
+	echo "    # $0 -d plex transmission"
+	echo "      Uninstall (DESTROY) plex and transmission"
+}
```
Ornias1993, line 98: This is a lot nicer indeed!

includes/global.sh
```diff
@@ -7,126 +7,144 @@
...
+		-e "s|^\($s\)\($w\)$s:$s\(.*\)$s\$|\1$fs\2$fs\3|p"  "${1}" |
+	awk -F$fs '{
+		indent = length($1)/2;
+		vname[indent] = $2;
+		for (i in vname) {if (i > indent) {delete vname[i]}}
+			if (length($3) > 0) {
+				vn=""; for (i=0; i<indent; i++) {vn=(vn)(vname[i])("_")}
+				printf("export %s%s%s=\"%s\"\n", "'$prefix'",vn, $2, $3);
+			}
+	}'
```
Ornias1993, line 22: The right indentation makes this function A LOT more readable! 👍

includes/global.sh
```diff
@@ -7,126 +7,144 @@
...
+	gitbranch=$(git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)")
+	branch=${gitbranch:-}
+
+	if [ "$(git config --get remote.origin.url)" = "https://github.com/Ornias1993/jailman" ]
 	then
-		echo "script requires update"
-		git reset --hard > /dev/null 2>&1
-		git pull > /dev/null 2>&1
-		echo "script updated, please restart the script manually"
+		echo "The repository has been moved, please reinstall using the new repository: jailmanager/jailman"
```
Ornias1993, line 34: We might want to set the current remote to the right URL instead....

includes/global.sh
```diff
@@ -7,126 +7,144 @@
...
 		exit 1
 	fi
-else
-	if ! iocage create -n "${1}" -p /tmp/pkg.json -r "${global_jails_version}" interfaces="${jailinterfaces}" ip4_addr="vnet0|${!jailip4}" defaultrouter="${!jailgateway}" vnet="on" allow_raw_sockets="1" boot="on" ${setextra} -b
-	then
-		echo "Failed to create jail"
-		exit 1
+
+	echo "Checking config..."
+	local blueprintpkgs blueprintports jailinterfaces jailip4 jailgateway jaildhcp setdhcp blueprintextraconf jailextraconf setextra reqvars reqvars
```
Ornias1993, line 70: Nice way of defining this, should've thought of this.

jailman.sh
asmod3us: missed these, looking...

jailman.sh
asmod3us: actually i should test that because of eval, but yes they bubble up to the error trap and we exit

includes/global.sh
asmod3us: i pulled my hair out for a while because a function was inadvertently re-defining blueprint and tripped up the caller's flow!

Ornias1993:
Rebased on current dev

Ornias1993:
Considering global.sh was getting HUGE with this rebase, I split the config...

asmod3us:
> Considering global.sh was getting HUGE with this rebase, I split the config...
Great! I'll continue with that - the strict mode will go into its own file too

asmod3us:
And thanks for the rebase!

includes/global_functions.sh
```diff
@@ -143,6 +146,8 @@ jailcreate() {
 }
 
 createmount() {
+	# as this function is called from blueprints we need to re-enable strict mode
+	strict::mode
```
Ornias1993, line 151: If you call strict on initblueprint, doesn't that also propagate to these function calls?

includes/blueprint_functions.sh
```diff
@@ -54,7 +57,13 @@ initblueprint() {
 export -f initblueprint
 
 cleanupblueprint() {
-	link_traefik="jail_${1}_link_traefik"
+	# as this function is called from blueprints we need to re-enable strict mode
+	strict::mode
```
Ornias1993, line 62: cleanup blueprint is not called from blueprint but from jailman.sh

includes/blueprint_functions.sh
```diff
@@ -48,8 +48,8 @@ initblueprint() {
 	fi
 
 	if [ -z "${ip4_addr}" ]; then
-		DEFAULT_IF=$(iocage exec "$1" route get default | awk '/interface/ {print $2}')
-		declare -g "jail_ip=$(iocage exec "$1" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')"
+		DEFAULT_IF=$(iocage exec "${jail_name}" route get default | awk '/interface/ {print $2}')
+		declare -g "jail_ip=$(iocage exec "${jail_name}" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')"
```
Ornias1993: Agreed, this was a reminant from how we previously did it at the end of the blueprint...

includes/global_functions.sh
```diff
@@ -66,7 +66,7 @@ jailcreate() {
 	jail=${1:?}
 	blueprint=${2:?}
 
-	if [ -z $jail ] || [ -z $blueprint ]; then
+	if [ -z "$jail" ] || [ -z "$blueprint" ]; then
```
Ornias1993, line 70: Strange normally shellcheck would start shouting about this.... :s

includes/blueprint_functions.sh
asmod3us: ah, i'll get rid of it here then

includes/global_functions.sh
asmod3us: it propagates to other functions, but a blueprint install.sh runs as a subshell, which then calls this function. in my tests this did not propagate into subshells, but needs further testing with exporting BASHOPTS

includes/blueprint_functions.sh
```diff
@@ -86,11 +86,17 @@ exitblueprint() {
...
+	setdhcp=${!jaildhcp}
 
 	# Check if the jail is compatible with Traefik and copy the right default-config for the job.
-	if [ -z "${link_traefik}" ] || [ -z "${ip4_addr}" ]; then
+	if [ -z "${link_traefik}" ]; then
 		echo "Traefik-connection not enabled... Skipping connecting this jail to traefik"
 	else
+		if [ -z "${ip4_addr}" ] && [ "${setdhcp}" == "on" ] && [ -n "${jail_ip}" ]; then
+			echo "Traefik-connection with DHCP requires that the assigned IP adddress stays the same!"
+		fi
```
Ornias1993: Firstoff
jaildhcp is not always set, because it's also the default of nothing is set.

It also breaks with a rule I made some time ago:
If it interconnects, it can't use DHCP.

The support burden for DHCP is quite bug tbh and it's a HUGE security risk to point traefik at an IP adress that might change.

includes/blueprint_functions.sh
```diff
@@ -92,6 +92,8 @@ exitblueprint() {
 	# Check if the jail is compatible with Traefik and copy the right default-config for the job.
 	if [ -z "${link_traefik}" ]; then
 		echo "Traefik-connection not enabled... Skipping connecting this jail to traefik"
+	elif [[ ${link_traefik} =~ .*[[:space:]]$ ]]; then
+		echo "Trailing whitespace in linked traefik jail '${link_traefik}'"
```
Ornias1993, line 94: Why is this required?
Are we going to test for all sorts of different typo's everywhere?

because this error can be made in just about all input reference tbh...

.github/workflows/shellcheck.yml
```diff
@@ -17,3 +17,5 @@ jobs:
       - uses: actions/checkout@v1
       - name: Shell Linter
         uses: azohra/shell-linter@v0.3.0
+        env:
+          SHELLCHECK_OPTS: -P SCRIPTDIR
```
Ornias1993, line 21: thanks

includes/blueprint_functions.sh
Ornias1993: If it is still strict, That should be done indeed  :)

Ornias1993: Note:
It's called after jail destruction, to cleanup any remaining traefik configs...

includes/blueprint_functions.sh
asmod3us: i think `jailcreate` ensures `setdhcp` is always set:
```
	if [ -z "${setdhcp}" ] && [ -z "${!jailip4}" ] && [ -z "${!jailgateway}" ]; then
		echo 'no network settings specified in config.yml, defaulting to dhcp="on"'
		setdhcp="on"
	fi
```

i understand we don't want to support dynamic DHCP, but DHCP with static ips based on MAC/client id is quite common. how about a global flag that explicitly states you have a such a setup, e.g. `static_dhcp: true`?

includes/blueprint_functions.sh
asmod3us: yes i think this would make sense globally

includes/blueprint_functions.sh
asmod3us: 👍

includes/global_functions.sh
asmod3us: ¯\_(ツ)_/¯

includes/global_functions.sh
```diff
@@ -113,13 +113,13 @@ jailcreate() {
...
 	then
-		if ! iocage create -n "${jail}" -p /tmp/pkg.json -r "${global_jails_version:?}" interfaces="${jailinterfaces}" dhcp="on" vnet="on" allow_raw_sockets="1" boot="on" ${setextra} -b
+		if ! iocage create -n "${jail}" -p /tmp/pkg.json -r "${global_jails_version:?}" interfaces="${jailinterfaces}" dhcp="on" vnet="on" allow_raw_sockets="1" boot="on" "${setextra}" -b
 		then
 			echo "Failed to create jail"
 			exit 1
 		fi
 	else
-		if ! iocage create -n "${jail}" -p /tmp/pkg.json -r "${global_jails_version}" interfaces="${jailinterfaces}" ip4_addr="vnet0|${!jailip4}" defaultrouter="${!jailgateway}" vnet="on" allow_raw_sockets="1" boot="on" ${setextra} -b
+		if ! iocage create -n "${jail}" -p /tmp/pkg.json -r "${global_jails_version}" interfaces="${jailinterfaces}" ip4_addr="vnet0|${!jailip4}" defaultrouter="${!jailgateway}" vnet="on" allow_raw_sockets="1" boot="on" "${setextra}" -b
```
Ornias1993: shellcheck seems to be a bit hit or miss lately... didn't get these warnings myself past few executions.

It seems it sometimes just randomly skips certain portions and starts complaining on later PR's... weird.

asmod3us: yes i'm also running into issues. it's confused in my editor about the includes now

includes/global_functions.sh
Ornias1993: If you call it in initblueprint, thats executed inside the install.sh subshell... afaik...

includes/global_functions.sh
Ornias1993: Includes require the path starting from project root...

asmod3us: yep, i've set `-P SCRIPTDIR` as option but doesn't seem to work

includes/global_functions.sh
asmod3us: yes that should work

includes/blueprint_functions.sh
Ornias1993: setdhcp is just overruled during jailcreate. ${setdhcp} is a local variable not global.

`how about a global flag that explicitly states you have a such a setup, e.g. static_dhcp: true`, in that cause users could just as well enter the IP anyway...

I think it makes things confusing too... we would create linked jails that would require fixed IP's and on the other hand have traefik that doesn't, but does require it to be static which it can't verify and might open up unwanted services to the outside world of done wrong.

I try to work hard towards a "set and forget" type solution and I think this is a "set and have a headach" type of idea...

Ornias1993: Simply put:
I'm currently not even considering allowing dhcp for: Mariadb, influxdb and unifi (is using poller, still need to add the requirement in code actually I think).

Why would I just for traefik?

Ornias1993: If you want to add it just for traefik, I would allow it... 
with setdhcp to "OVERRIDE" (keep it vague)

But it needs a VERY clear warning in the docs like:
`If you set DHCP to override, we do not check if you have a static IP in some cases. This means you might break things and you will **NOT** get support if you have set this flag anywhere.`

asmod3us: for me, the headache is administrating static ips, and in lots of places. making the decision in one place, i.e. on the router is straightforward to me.

i want to use jailman with this setup and i'm aware of the pitfalls. what do you propose?

includes/blueprint_functions.sh
Ornias1993: I see no place anywhere where one would want to end with a space... :')

includes/blueprint_functions.sh
Ornias1993: See my edited comment.

asmod3us: that sounds cool, i'll look into that! would the `dhcp: override` live in the traefik config block or in the jail?

Ornias1993: Well we are talking about the jail creating it's own dynamic config. Traefik has nothing to do with it besides "eating" said config.

So I think setting it on a per-jail basis would be prefered and would also make it useable for other blueprints in the future

includes/global_functions.sh
Ornias1993: Remove that. It wasn't needed in the past either.

Ornias1993: Looked into it a bit, you know it would help if all new SH scripts where executable and such ;)

Ornias1993: (pushed a fix for that)

Ornias1993:
@asmod3us After setting them to be executable it's passing tests.

includes/global_functions.sh
asmod3us: ahhh that makes sense 👍

includes/blueprint_functions.sh
asmod3us: Agreed, this is now done on all config!

Ornias1993:
@asmod3us besides leading and trailing spaces, we might also want to forbid `|` entirely. It could create all sorts of issues in our sed commands and it isn't needed at all in any config parameter

includes/global_functions.sh
```diff
@@ -50,6 +50,24 @@ validate_config() {
...
+
+	cfg=$(parse_yaml "${SCRIPT_DIR}/includes/global.yml")
+	validate_config "${SCRIPT_DIR}/includes/global.yml" "$cfg"
+	eval "$cfg"
+
+	cfg=$(parse_yaml "${SCRIPT_DIR}/config.yml")
+	validate_config "${SCRIPT_DIR}/config.yml" "$cfg"
+	eval "$cfg"
+}
+
```
Ornias1993, line 70: Maybe just seperate all functions that are only used in jailman.sh to `preload_functions.sh`?

Ornias1993: `gitupdate`, `load_config`, `validate_config`

includes/blueprint_functions.sh
asmod3us: So full example:

config.yml:
```
...
  sabnzbd:
    blueprint: sabnzbd
    dhcp: override
    domain_name: my.cool.domain
    link_traefik: traefik

  traefik:
    blueprint: traefik
    dashboard: true
    domain_name: traefik.nyx.io
    dns_provider: cloudflare
    cert_email: fake@email.net
    cert_wildcard_domain: nyx.io
    cert_env:
...
```
Is that what you propose?

Couple of things:
- just setting `dhcp: on` works with traefik setup and jail creation in the current codebase and my proposed change. traefik just grabs the current ip of the jail and that'

- setting dhcp to override means adjusting the jail creation code to figure out what to do with 'override' (mostly just rewrite it to on and print a warning?).

includes/global_functions.sh
asmod3us: how about `libjailmain.sh`? :-)

asmod3us: > @asmod3us besides leading and trailing spaces, we might also want to forbid `|` entirely. It could create all sorts of issues in our sed commands and it isn't needed at all in any config parameter

Sure, adding!

Ornias1993: I don't get what it's function would be with libjailmain.sh.
Maybe preload is the wrong them.

How about:
init_functions.sh

Those are all functions that are basically run only to fetch/process the basic data structure and/or prepare jailman for a run...

Ornias1993:
Rebased with latest changes

includes/global_functions.sh
asmod3us: naming things is hard :) i defer to your judgement

includes/blueprint_functions.sh
Ornias1993: > Is that what you propose?

Yeah that was what I proposed config wise.







> * just setting `dhcp: on` works with traefik setup and jail creation in the current codebase and my proposed change. traefik just grabs the current ip of the jail and that'

`dhcp: on` should only work on jail creation, `dhcp: override` enables it to be used in traefik too.



> * setting dhcp to override means adjusting the jail creation code to figure out what to do with 'override' (mostly just rewrite it to on and print a warning?).

Something like this yes (there are multiple ways of doing this in jailcreate)...

includes/global_functions.sh
Ornias1993: @asmod3us 
Hard and there is no "right" ways really... Only a "consistent" ways and considering I named most things it's not surprising my way is the most consistent... That doesn't make it beter ;)

includes/blueprint_functions.sh
asmod3us: ok, got it. so for jail creation we just parse override as equal to on and let traefik proceed only with override. i'll implement this

Ornias1993:
@asmod3us You might also want to check out #13 while you are at it...

asmod3us:
@Ornias1993 for the traefik setup, is it intentional that the toml files get changed after being copied over to traefik /dynamic dir?

E.g.:

```
...
sed -i '' "s|placeholderusers|${users//&/\\&}|" /mnt/"${global_dataset_config}"/"${link_traefik}"/temp/"${jail_name}"_auth_basic.toml
mv /mnt/"${global_dataset_config}"/"${link_traefik}"/temp/"${jail_name}"_auth_basic.toml /mnt/"${global_dataset_config}"/"${link_traefik}"/dynamic/"${jail_name}"_auth_basic.toml
sed -i '' "s|\"retry\"|\"retry\",\"${1//&/\\&}-basic-auth\"|" /mnt/"${global_dataset_config}"/"${link_traefik}"/temp/"${jail_name}".toml
```

Ornias1993:
@asmod3us Shoot. No ofc not >.<

asmod3us:
> @asmod3us Shoot. No ofc not >.<

Fixing... also simplifying the seds a bit

asmod3us:
> > @asmod3us Shoot. No ofc not >.<
> 
> Fixing... also simplifying the seds a bit

Oh it's actually two different files (`jail.toml` and `jail_auth_something.toml`). Is the intended outcome to have one toml in traefik or two?

Ornias1993:
> > > @asmod3us Shoot. No ofc not >.<
> > 
> > 
> > Fixing... also simplifying the seds a bit
> 
> Oh it's actually two different files (`jail.toml` and `jail_auth_something.toml`). Is the intended outcome to have one toml in traefik or two?

Sorry i was just eating and didn't read.
I was wrong, this is all how it is supposed to work, please don't mess with it too much if you don't precisely figured out how it is supposed to work. It's hell to get right.

asmod3us:
> > > > @asmod3us Shoot. No ofc not >.<
> > > 
> > > 
> > > Fixing... also simplifying the seds a bit
> > 
> > 
> > Oh it's actually two different files (`jail.toml` and `jail_auth_something.toml`). Is the intended outcome to have one toml in traefik or two?
> 
> Sorry i was just eating and didn't read.
> I was wrong, this is all how it is supposed to work, please don't mess with it too much if you don't precisely figured out how it is supposed to work. It's hell to get right.

Right, so it's two files, and one is being put into place later. I think I understand it now :)

Ornias1993:
To be clear: Not to shit on you or anything, but if you don't understand how changes in those files, affect the Traefik config it gets hairy to change things. Because their manual is kinda mediocre...

What is created here are dynamic config files, but they both contain different config. One contains config for just connecting the jail and the other contains config for what they call "middleware" which are basically plugins. 

So one file gets copied over to the dynamic config folder and gets loaded, loading the middleware "plugins" and the other gets copied loading the actual jail, which references the middleware that was created earlier.

It doesn't have to be in this order, but it works out to be cleaner and less error prone imho.

asmod3us:
That dynamic config is brilliant. Not looking to change these files at all, just combining the sed commands into one

Ornias1993:
> That dynamic config is brilliant. Not looking to change these files at all, just combining the sed commands into one

That might not be the best of idea for some of those seds, as there might be additional middlewares and config files being added later...

asmod3us:
Ran into a weird issue with basic auth credentials: my test password included `$2` and that gets replaced to whatever is in $2 at the time by the yaml parser. Or it craps out about it not being defined.

Ornias1993:
@asmod3us good catch.
I don't know if thats fixable.

We could just work with plaintext passwords and hash them on demand in the script?

asmod3us:
That's an idea!. The other thing that would work is to have the awk section of parse_yaml escape everything for the shell. Or find a proper yaml parser for the shell 🙈

Ornias1993:
@asmod3us actually... I learned jq is natively included in FreeNAS lately.
So you could rewrite it to use jq if you like...

asmod3us:
Pushed the changes around traefik cfg handling. I validated that the resulting traffic config without and with basic auth is exactly the same as before.

asmod3us:
> @asmod3us actually... I recently learned jq is natively included in FreeNAS.
> So you could rewrite it to use jq if you like...

Does jq parse YAML?

Ornias1993:
> > @asmod3us actually... I recently learned jq is natively included in FreeNAS.
> > So you could rewrite it to use jq if you like...
> 
> Does jq parse YAML?

I'm an idiot. nvm... 
Doing three things at once xD

asmod3us:
> > > @asmod3us actually... I recently learned jq is natively included in FreeNAS.
> > > So you could rewrite it to use jq if you like...
> > 
> > 
> > Does jq parse YAML?
> 
> I'm an idiot. nvm...
> Doing three things at once xD

No, i think there is merit to it.

This python snippet runs on FreeNAS:
```
import yaml
import json

with open('./config.yml') as f:
        print(json.dumps(yaml.safe_load(f)));
```

and it produces json from our `config.yml` which could then be queried with `jq`.

Ornias1993:
@asmod3us I suggest keeping that a seperate PR, after this is done.

asmod3us:
Yep, this was just a quick experiment.

Ornias1993:
> Yep, this was just a quick experiment.

Still its something to think about, considering we also want to go support lists...

asmod3us:
> > Yep, this was just a quick experiment.
> 
> Still its something to think about, considering we also want to go support lists...

How do you feel about switching to python as a language?

Ornias1993:
@asmod3us Yes and No? :P
I don't hate python, but I feel it's risky to rely on python being included in FreeNAS in the future.

Ornias1993:
However, all things considered:
If it's for just the core functions, i'm interested.... because it makes the parsing of all those things a lot cleaner and more code efficient.

asmod3us:
> @asmod3us Yes and No? :P
> I don't hate python, but I feel it's risky to rely on python being included in FreeNAS in the future.
> 

Isn't TrueNAS/FreeNAS Core written in python and therefore a safe bet?

asmod3us:
Definitely food for thought, in this PR i'm going to keep it bashy ;)

asmod3us:
I'm going to mark this as ready for review, I've been running with this PR for a couple of days and it improves QoL a lot.

Ornias1993:
> Definitely food for thought, in this PR i'm going to keep it bashy ;)

Yeah I think it's a three step proces: 
1. Good bash
2. Replacing complex functions with a "real" programming language
3. Replacing the remaining code with a real programming language.

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
+#!/usr/local/bin/bash
+# shellcheck disable=SC1003
+
+# shellcheck source=libstrict.sh
+source "${SCRIPT_DIR}/includes/libstrict.sh"
+strict::mode
```
Ornias1993, line 6: Is this needed here?

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
...
+			do
+				linkvalue="jail_${val}_${linkvar}"
+				linkval=${!linkvalue:-}
+				declare -g "${var}_${linkvar}=${linkval}"
+			done
+		fi
+	done
+
+	declare -g "jail_root=/mnt/${global_dataset_iocage}/jails/${jail_name}/root"
+	declare -g "blueprint_dir=${SCRIPT_DIR}/blueprints/${!blueprint}"
```
Ornias1993, line 39: Ill make sure those are used in all the current code...

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
...
+		declare -g "jail_ip=$(iocage exec "${jail_name}" ifconfig "$DEFAULT_IF" | awk '/inet/ { print $2 }')"
+	else
+		declare -g "jail_ip=${ip4_addr%/*}"
+	fi
+}
+export -f initblueprint
+
+cleanupblueprint() {
+	local jail_name=${1:?}
+	local status_message=${2:-}
```
Ornias1993, line 63: Why is this needed?

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
...
+		rm -f /mnt/"${global_dataset_config}"/"${link_traefik}"/dynamic/"${jail_name}"_auth_basic.toml
+		rm -f /mnt/"${global_dataset_config}"/"${link_traefik}"/dynamic/"${jail_name}"_auth_forward.toml
+	fi
+}
+export -f cleanupblueprint
+
+exitblueprint() {
+	# as this function is called from blueprints we need to re-enable strict mode
+	source "${SCRIPT_DIR}/includes/libstrict.sh"
+	strict::mode
```
Ornias1993, line 79: exit blueprint is called from the install.sh subshell, which has already always strictmode enabled on initblueprint, right?

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
...
+
+	traefik_root=/mnt/"${global_dataset_config}"/"${link_traefik}"
+	traefik_tmp=${traefik_root}/temp
+	traefik_dyn=${traefik_root}/dynamic
+
+	# Check if the jail is compatible with Traefik and copy the right default-config for the job.
+	if [ -z "${link_traefik}" ]; then
+		echo "Traefik-connection not enabled... Skipping connecting this jail to traefik"
+	elif { [ -z "${setdhcp}" ]  || [ "${setdhcp:-}" == "on" ]; }  && [ -z "${!jailip4:-}" ] && [ -z "${!jailgateway:-}" ]; then
+		echo "Traefik requires dhcp: override in order to work with DHCP."
```
Ornias1993, line 104: Maybe add a short format of the Readme warning here too...

includes/blueprint_functions.sh
```diff
@@ -0,0 +1,192 @@
...
+			traefik_status="success"
+		elif [ -n "${traefik_auth_forward}" ]; then
+			echo "Adding forward auth to Traefik for jail ${jail_name}"
+			cp "${traefik_includes}/default_auth_forward.toml" "${traefik_tmp}/${jail_name}_auth_forward.toml"
+			sed -i '' \
+				-e "s|placeholdername|${1//&/\\&}|" \
+				-e "s|placeholderauthforward|${traefik_auth_forward//&/\\&}|" \
+				"${traefik_tmp}/${jail_name}_auth_forward.toml"
+			mv "${traefik_tmp}/${jail_name}_auth_forward.toml" "${traefik_dyn}/${jail_name}_auth_forward.toml"
+			sed -i '' "s|\"retry\"|\"retry\",\"${1//&/\\&}-forward-auth\"|" "${traefik_tmp}/${jail_name}.toml"
```
Ornias1993, line 167: Nice sed cleanup! :)

includes/global_functions.sh
```diff
@@ -0,0 +1,127 @@
...
+	jailinterfaces="jail_${jail}_interfaces"
+	jailip4="jail_${jail}_ip4_addr"
+	jailgateway="jail_${jail}_gateway"
+	jaildhcp="jail_${jail}_dhcp"
+	setdhcp=${!jaildhcp}
+	blueprintextraconf="blueprint_${blueprint}_custom_iocage"
+	jailextraconf="jail_${jail}_custom_iocage"
+	setextra="${!blueprintextraconf:-}${!jailextraconf:+ ${!jailextraconf}}"
+	reqvars=blueprint_${blueprint}_reqvars
+	reqvars="${!reqvars:-}${global_jails_reqvars:+ ${!global_vars_reqvars}}"
```
Ornias1993, line 33: I get your python comment now. This is becomming a clusterduck.

includes/global_functions.sh
```diff
@@ -0,0 +1,127 @@
...
+		else
+			echo "Dataset already exists, skipping creation of ${dataset}"
+		fi
+
+		if [ -n "${jail}" ] && [ -n "${mountpoint}" ]; then
+			iocage exec "${jail}" mkdir -p "${mountpoint}"
+			if [ -n "${fstab}" ]; then
+				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" "${fstab}" || exit 1
+			else
+				iocage fstab -a "${jail}" /mnt/"${dataset}" "${mountpoint}" nullfs rw 0 0 || exit 1
```
Ornias1993, line 118: You might want another error catch here instead of my old `exit 1`?

includes/init_functions.sh
```diff
@@ -0,0 +1,107 @@
...
+	local s='[[:space:]]*' w='[a-zA-Z0-9_]*' fs=$(echo @|tr @ '\034')
+	sed -ne "s|^\($s\)\($w\)$s:$s\"\(.*\)\"$s\$|\1$fs\2$fs\3|p" \
+		-e "s|^\($s\)\($w\)$s:$s\(.*\)$s\$|\1$fs\2$fs\3|p"  "${1}" |
+	awk -F$fs '{
+		indent = length($1)/2;
+		vname[indent] = $2;
+		for (i in vname) {if (i > indent) {delete vname[i]}}
+			if (length($3) > 0) {
+				vn=""; for (i=0; i<indent; i++) {vn=(vn)(vname[i])("_")}
+				printf("export %s%s%s=\"%s\"\n", "'$prefix'",vn, $2, $3);
```
Ornias1993, line 23: The right indent really cleans this up...

includes/init_functions.sh
```diff
@@ -0,0 +1,107 @@
...
+	done
+
+	cfg=$(parse_yaml "${SCRIPT_DIR}/includes/global.yml")
+	validate_config "${SCRIPT_DIR}/includes/global.yml" "$cfg"
+	eval "$cfg"
+
+	cfg=$(parse_yaml "${SCRIPT_DIR}/config.yml")
+	validate_config "${SCRIPT_DIR}/config.yml" "$cfg"
+	eval "$cfg"
+}
```
Ornias1993, line 71: Seperating this really cleans things up a bit...

Ornias1993:
@disconn3ct
As you requested a safer use of bash, you might want to review this...

includes/blueprint_functions.sh
asmod3us: Which part specifically? the local declarations?

includes/blueprint_functions.sh
asmod3us: Investigating that...

includes/blueprint_functions.sh
asmod3us: Hmm maybe we make the naming consistent and i call it `blueprint_root`?

includes/blueprint_functions.sh
Ornias1993: I don't think there is a ${2} with cleanupblueprint ;)

includes/blueprint_functions.sh
asmod3us: You're right yes. However I'm still a little unclear if the strict mode transports to function calls when they're exported by the parent shell. We can remove it and see if it works.

includes/blueprint_functions.sh
Ornias1993: No I don't think it is inconsistant.

jail root point to the actual root folder of the jail (named root, mounted as root (/) in the jail). Include_dir and blueprint_dir point to directories (dir) inside the jailman folder structure.

includes/blueprint_functions.sh
Ornias1993: Seems like an easy test:
- Remove
- create buildin-error
- run.

includes/blueprint_functions.sh
asmod3us: Yes I will include a pointer to the readme.

includes/blueprint_functions.sh
asmod3us: ah ok, perfect!

includes/blueprint_functions.sh
asmod3us: ah good catch, out it goes

includes/blueprint_functions.sh
```diff
@@ -106,7 +106,9 @@ exitblueprint() {
 		echo "Trailing whitespace in linked traefik jail '${link_traefik}'"
 	else
 		if [ -z "${ip4_addr}" ] && [ "${setdhcp}" == "override" ] && [ -n "${jail_ip}" ]; then
-			echo "Traefik with DHCP requires that the jail's assigned IP adddress stays the same!"
+			echo "Warning: Traefik with DHCP requires that the jail's assigned IP adddress stays the same."
+			echo "Please see the README for details. If you cannot guarantee static IP assignment this will"
+			echo "BREAK in unexpected ways and potentially expose services you did not intend to expose."
```
Ornias1993, line 110: We might also use something like this on row 104:
`echo "Traefik requires dhcp: override in order to work with DHCP."`

To prevent people from actually turning it on... Now we warn them after the fact.

asmod3us: Pointing to README now

includes/global_functions.sh
asmod3us: Failing with error message now.

includes/blueprint_functions.sh
asmod3us: Seems it's not needed - addressed.

asmod3us:
> Passing shellcheck is required.

Oh it's unhappy about following paths again :(

PR #173, opened 2020-06-02
syntax check blueprint install script before creating jail

asmod3us:


asmod3us:
This should catch more issues with install scripts.

Ornias1993:
Yes indeed, adding right away... :)
Thanks!

PR #179, opened 2020-06-12
Fix Unifi DB creation in InfluxDB

### LINKED ISSUE #151: Unifi database not created in InfluxDB
# Issue Template

## Description
When creating the Unifi jail w/ integrated Controller, the unifi database is not being created in Influx.

### Steps to Reproduce

1. Create Influx jail and Unifi Jail 
2.  Observe no unifi database being created
3.  

Noticed this when testing Grafana. I'll take a look when I get to the Unifi fixes.

colemamd:
# Pull Request Template

### Purpose
Fixes #151 <br>

### Notes:
Contains only the fix, no updates for 1.3

- [x] Code is sufficiently commented
- [x] Code is indented with tabs and not spaces
- [x] The PR has been tested
- [x] Any new files are named using lowercase (to avoid issues on case sensitive file systems)

### DIFF
--- a/blueprints/unifi/install.sh
+++ b/blueprints/unifi/install.sh
@@ -63,7 +63,7 @@ iocage exec "${1}" sysrc unifi_enable=YES
 iocage exec "${1}" service unifi start
 
 # shellcheck disable=SC2154
-if [[ ! "${!POLLER}" ]]; then
+if [[ -z "${POLLER}" ]]; then
   echo "Installation complete!"
   echo "Unifi Controller is accessible at https://${JAIL_IP}:8443."
 else
--- a/blueprints/unifi/update.sh
+++ b/blueprints/unifi/update.sh
@@ -2,7 +2,7 @@
 # This file contains the update script for unifi
 # Unifi Controller is updated through pkg, Unifi-Poller is not. This script updates Unifi-Poller
 
-if [[ ! "${!POLLER}" ]]; then
+if [[ -z "${POLLER}" ]]; then
 	echo "Skipping Unifi Poller for update, not installed"
 else
 	


blueprints/unifi/update.sh
```diff
@@ -2,7 +2,7 @@
 # This file contains the update script for unifi
 # Unifi Controller is updated through pkg, Unifi-Poller is not. This script updates Unifi-Poller
 
-if [[ ! "${!POLLER}" ]]; then
+if [[ -z "${POLLER}" ]]; then
```
Ornias1993, line 5: Are you VERY sure? I'm sure this is wrong.

removing the ! is new syntax in de dev branch, it's not minor-dev compatible.

colemamd: yeah, this quick-fix just checks if there is a poller variable, it does not need expansion. I'll come up with a better way (i.e. should be able to put 'false')

Ornias1993: You cant use POLLER as variable in update, it's undefined.
Be aware: this is minor-dev NOT dev.

Minor dev does NOT use the new way of defining variables. so you NEED to use ${!VARIABLE} AND the variable needs to be defined in the script.

Remember this is version 1.2 still, not version 1.3!

Ornias1993: I'm not talking about changing this check, i'm saying ${POLLER} is undefined here.

blueprints/unifi/install.sh
```diff
@@ -63,7 +63,7 @@ iocage exec "${1}" sysrc unifi_enable=YES
 iocage exec "${1}" service unifi start
 
 # shellcheck disable=SC2154
-if [[ ! "${!POLLER}" ]]; then
+if [[ -z "${POLLER}" ]]; then
```
Ornias1993, line 66: `echo ${POLLER}`
Would result in:
`jail_JAILNAME_unifi_poller`

Remember, this is minor-dev, so its still v1.2.x, NOT v1.3.x

Correct syntax is really:
${!POLLER}

Ornias1993:
@colemamd I think i'm getting confused by this.
I understoot from you this issue existed in MASTER (THe current latest release), but all your fixes are for dev.

If this is meant for dev, please send it to the devbranch and clearly state so in your bugreport.
If this is meant for master, it needs to be a fix for a bug in master and send to minor-dev.

Is this a bug in master or dev? I don't get it anymore.

colemamd:
It's for master and am working on updating it for inclusion in minor-dev

Ornias1993:
@colemamd Thanks for clearing that up.
Your current PR doesn't work, I can assure you that.

Hence I'll close it, feel free to reopen if it at least works. (this can't)

PR #209, opened 2020-07-24
Cleanup and Simplification

### LINKED ISSUE #14: Summary report printout
It would be nice to have a consolidated log and passwords file created on each run.

### LINKED ISSUE #122: Add an destructive removal option
# Issue Template
### Detailed Bug Report
It might be needed sometimes (for example: when the first install failed halfway through) to fully remove said jail + user files/config.

It would be nice to add an option for this:
`-df`

When set it would require the user to enter a phrase to continue:
"I REALLY WANT TO DESTROY ALL MY DATA FROM $JAILNAME"

### LINKED ISSUE #204: Move internal config files to JSON
While YAML is still prefered for user-facing config files, we should use a more "structure" focussed config format ourselves. JSON in this case. Its a little more picky and our automation can easily verify it using a schema, which would increase code quality.

### LINKED ISSUE #205: remove "jail_" from config variables
Putting jails_  variables was introduced when we seperated them from blueprint variables.
Currently blueprints variables are being removed and merged with the plugin manifest JSON.

So there is not reason to keep this prefix.

Ornias1993:
This PR cleans up and simplifies some portions of the codebase.

- remove jail_ prefix (Fixes #205 )
- incremend IP-adresses in config.yml
- Fix ports, 
- remove unneeded whitespace, 
- remove plugin_extra_conf (not possible anymore)
- removes explicity dhcp setting
- removes override dhcp setting
- adds/moves warning in case dhcp is used
- Removes redundant whitespace-after-variable check
- Adds shortcuts for echo colors
- Move global config to json file (same structure as manifests) (Fixes #204 )
- Add config removal option (Fixes #122 )
- Add logging to /var/log/jailman.log
- Add summary generation (Fixes #14)
- Add automatically creating /root/PLUGIN_INFO for plugins being installed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /config.yml
 /tmp/
+/summaries/
--- a/config.yml.example
+++ b/config.yml.example
@@ -11,135 +11,133 @@ global:
     # (is going to contain: Complete and Incomplete datasets for downloads)
     downloads: tank/downloads
 
-jail:
-  plexjail:
-    blueprint: plex
-    ip4_addr: 192.168.1.99/24
-    gateway: 192.168.1.1
-    beta: false
+plexjail:
+  blueprint: plex
+  ip4_addr: 192.168.1.200/24
+  gateway: 192.168.1.1
 
-  lidarrjail:
-    blueprint: lidarr
-    ip4_addr: 192.168.1.99/24
-    gateway: 192.168.1.1
+lidarrjail:
+  blueprint: lidarr
+  ip4_addr: 192.168.1.201/24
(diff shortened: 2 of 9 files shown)


Ornias1993:
@asmod3us I also followed your previous advice and simpified DHCP.
Although we warn about it when used, thats about it and is now seperated from the Traefik code....

Ornias1993:
I'm going to quickly merge this... Its tested and will send in bugfixes if anything breaks.
This is basically (besides traefik) the last non-python rework/refactorI expect to be done for 1.3.0 or 1.4.0

